PR #7, opened 2017-11-09
PP-2808 Upgrade Dropwizard and other libraries

DanailMinchev:
## WHAT

- Added `dependencyManagement` for Dropwizard
- Upgraded Dropwizard to latest stable version (`1.2.0`) and upgraded all other libraries

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
 
 
     <properties>
-        <dropwizard.version>1.0.6</dropwizard.version>
+        <dropwizard.version>1.2.0</dropwizard.version>
         <mockserver.version>3.10.4</mockserver.version>
         <jersey2.version>2.25.1</jersey2.version>
         <hamcrest.version>1.3</hamcrest.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -1,40 +1,63 @@
...
+    <prerequisites>
+        <maven>3.0.0</maven>
+    </prerequisites>
+
     <groupId>uk.gov.pay</groupId>
-    <version>0.1-SNAPSHOT</version>
     <artifactId>pay-direct-debit-connector</artifactId>
-
+    <version>0.1-SNAPSHOT</version>
+    <packaging>jar</packaging>
```
simad: Do we need this? We are not really packaging dd connector (or any java app) as a jar

pom.xml
```diff
@@ -1,40 +1,63 @@
...
+        <guava.version>23.3-jre</guava.version>
+        <hamcrest.version>1.3</hamcrest.version>
+        <rest-assured.version>3.0.5</rest-assured.version>
     </properties>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>io.dropwizard</groupId>
+                <artifactId>dropwizard-bom</artifactId>
```
simad: nice!

pom.xml
```diff
@@ -45,22 +68,12 @@
         <dependency>
             <groupId>org.liquibase</groupId>
             <artifactId>liquibase-core</artifactId>
-            <version>3.3.5</version>
+            <version>${liquibase.version}</version>
```
simad: I see that we are excluding 
```
    <exclusions>
                <exclusion>
                    <groupId>javax.servlet</groupId>
                    <artifactId>jstl</artifactId>
                </exclusion>
```
from liquibase-core in cc connector. Have you checked if we still need to exclude that?

pom.xml
DanailMinchev: I thought we are using the generated `jar` here:
- https://github.com/alphagov/pay-direct-debit-connector/blob/f3830c54a32d427df77341ccfb9bd108e3c137e1/Dockerfile#L18
- https://github.com/alphagov/pay-direct-debit-connector/blob/f3830c54a32d427df77341ccfb9bd108e3c137e1/docker-startup.sh#L3
Or you mean something else? Thanks

simad: I was just wondering if it was required to put it in the pom, because I think it's the default packaging if not specified. After clarifying you're doing it to make an implicit explicit, i'd say you can leave it there 👍

pom.xml
DanailMinchev: I haven't check that, thanks.
So, running `mvn dependency:tree -Dverbose` and searching for `jstl` I was unable to find anything.
We have only `javax.servlet:javax.servlet-api:jar:3.1.0:compile`.
I checked in cc connector as well: there is no `jstl` dependency.
As discussed may be this dependency was in previous versions, that is why I will skip adding it as an `exclusion`
Thanks

DanailMinchev:
Thanks @simad 
As discussed with you and @rauligs I will merge this version (with latest Jackson version `2.9.2` explicitly defined in `pom.xml` which overrides the Dropwizard's Jackson version `2.9.1`).
We can integrate `maven enforcer plugin` as you mentioned if needed.

PR #8, opened 2017-11-14
Fix EclipseLink and java.lang.SecurityException

DanailMinchev:
## WHAT

The eclipselink 2.7.0 artifact depends on javax.persistence 2.2.0 and both contain classes that are in the package javax.persistence. However the javax.persistence jar is signed while the eclipselink one is not. So the Java runtime will complain, when loading a class from the package javax.persistence in the eclipselink jar, that it's lack of signing doesn't match with classes already loaded from the same package in the javax.persistence jar.

When running the app, we are receiving:
```
java.lang.SecurityException: class "javax.persistence.PersistenceUtil"'s signer information does not match signer information of other classes in the same package
```

More information:

- https://bugs.eclipse.org/bugs/show_bug.cgi?id=525457
- https://stackoverflow.com/questions/45870753/eclipselink-2-7-0-and-jpa-api-2-2-0-signature-mismatch

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,12 +79,29 @@
             <groupId>org.eclipse.persistence</groupId>
             <artifactId>eclipselink</artifactId>
             <version>${eclipselink.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.eclipse.persistence</groupId>
+                    <artifactId>javax.persistence</artifactId>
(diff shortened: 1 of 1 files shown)


DanailMinchev:
Thanks, yes, we can introduce `maven enforcer plugin` as you mentioned earlier. We should discuss this.

PR #14, opened 2017-11-16
PP-2808 Add connection to Direct Debit Database

rauligs:
## WHAT
- Application startup wait for database to be available.

- Add Database connectivity check into Healthcheck

- Add SSL Configuration


pom.xml
```diff
@@ -25,10 +25,11 @@
         <eclipselink.version>2.7.0</eclipselink.version>
         <jackson.version>2.9.2</jackson.version>
         <guice.version>4.1.0</guice.version>
-        <guava.version>23.3-jre</guava.version>
+        <guava.version>21.0</guava.version>
```
DanailMinchev: Can we use latest version here `23.4-jre`, it was `23.3-jre`, not sure why we need `21.0`?

pom.xml
```diff
@@ -164,6 +132,18 @@
...
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>${mockito.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.spotify</groupId>
+            <artifactId>docker-client</artifactId>
+            <version>8.1.1</version>
```
DanailMinchev: Can we move `version` to `properties` as well? The latest version is `8.9.2` can we use that?

src/test/java/uk/gov/pay/directdebit/infra/PostgresContainer.java
```diff
@@ -0,0 +1,127 @@
...
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.util.List;
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+
+import static java.lang.Integer.parseInt;
+import static java.util.stream.Collectors.joining;
+
+public class PostgresContainer {
```
DanailMinchev: Great idea! 👍

pom.xml
rauligs: Fixed

pom.xml
rauligs: Updated

src/test/java/uk/gov/pay/directdebit/infra/PostgresContainer.java
rauligs: Is actually copied from somewhere! adminusers or connector

PR #19, opened 2017-12-01
PP-2971 Add PaymentRequest and Token entities, with relative DAOs

simad:
This PR introduces two new entities, `PaymentRequest` and `Token`, with relative DAOs and test infrastructure.

- `PaymentRequest` and `Token` are modelled according to https://payments-platform.atlassian.net/browse/PP-2841. There is a difference: `payment_provider` is not in the`payment_requests` table, as will be inferred from the gateway account once that is modelled
- The DAOs are built according to JDBI 3 Declarative APIs
- Migrations are currently run before each test (See the `DAOItestBase`'s rules), this will changed once the DB spike is completed and the work is brought in master

### DIFF
--- a/src/main/resources/migrations/00001_create_table_payment_requests.sql
+++ b/src/main/resources/migrations/00001_create_table_payment_requests.sql
@@ -1,0 +1,16 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_table-payment_requests
+CREATE TABLE payment_requests (
+    id BIGSERIAL PRIMARY KEY,
+    external_id VARCHAR(26),
+    gateway_account_id BIGSERIAL NOT NULL,
+    amount BIGINT NOT NULL,
+    reference VARCHAR(255) NOT NULL,
+    description VARCHAR(255) NOT NULL,
+    return_url TEXT NOT NULL,
+    created_date TIMESTAMP WITH TIME ZONE DEFAULT (now() AT TIME ZONE 'utc') NOT NULL,
+    version INTEGER DEFAULT 0 NOT NULL
+);
+--rollback drop table payment_requests;
+
--- a/src/test/java/uk/gov/pay/directdebit/infra/DaoITestBase.java
+++ b/src/test/java/uk/gov/pay/directdebit/infra/DaoITestBase.java
@@ -1,0 +1,68 @@
+package uk.gov.pay.directdebit.infra;
+
+import io.dropwizard.jdbi.OptionalContainerFactory;
+import liquibase.Liquibase;
+import liquibase.database.jvm.JdbcConnection;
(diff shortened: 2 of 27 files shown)


src/main/resources/migrations/00001_create_table_payment_requests.sql
```diff
@@ -0,0 +1,16 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_table-payment_requests
+CREATE TABLE payment_requests (
+    id BIGSERIAL PRIMARY KEY,
```
rauligs: To use Big ints and sequence generators that allows preallocation

rauligs: Applies for all entities

src/test/java/uk/gov/pay/directdebit/infra/DaoITestBase.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.skife.jdbi.v2.DBI;
+import uk.gov.pay.directdebit.util.DatabaseTestHelper;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.util.Properties;
+
+public class DaoITestBase {
```
rauligs: This can be refactored afterwards too

src/test/java/uk/gov/pay/directdebit/util/DatabaseTestHelper.java
```diff
@@ -0,0 +1,82 @@
...
+
+import org.skife.jdbi.v2.DBI;
+import org.skife.jdbi.v2.util.StringColumnMapper;
+import uk.gov.pay.directdebit.payments.fixtures.PaymentRequestFixture;
+import uk.gov.pay.directdebit.payments.fixtures.TokenFixture;
+
+import java.sql.Timestamp;
+import java.util.Map;
+
+public class DatabaseTestHelper {
```
rauligs: Fixtures looks very nice! but would be even nicer for them to insert themselves so only needed JDBI and not databasehelper (as done in new end to end approach) and therefore keeping everything small and much easily localisable.

src/main/resources/migrations/00001_create_table_payment_requests.sql
simad: :+1: will change as part of the next PR

src/test/java/uk/gov/pay/directdebit/util/DatabaseTestHelper.java
simad: 👍  Will change too as part of the next PR

PR #23, opened 2017-12-01
PP-2971 Add state machine

simad:
- The new state machine is represented as a directed value graph. A node is a `PaymentState`, an edge connecting two nodes is an event, which represent a transition between the states.
- If there's an invalid state transition we are returning an `Internal server error`, mapped using the dropwizard exception mapper


src/main/java/uk/gov/pay/directdebit/payments/model/PaymentStatesGraph.java
```diff
@@ -0,0 +1,81 @@
...
+
+        return ImmutableValueGraph.copyOf(graph);
+    }
+    private final ImmutableValueGraph<PaymentState, PaymentRequestEvent.SupportedEvent> goCardlessGraphStates;
+
+    private PaymentStatesGraph() {
+        this.goCardlessGraphStates = buildGoCardlessStatesGraph();
+    }
+
+    public PaymentState getNextValidStateForEvent(PaymentState from, SupportedEvent event) {
```
heathd: not sure that the word `Valid` adds anything here?

how about `PaymentState getNextStateForEvent(PaymentState from, SupportedEvent event)`

or even `PaymentState next(PaymentState from, SupportedEvent event)`

src/test/java/uk/gov/pay/directdebit/payments/model/PaymentStatesGraphTest.java
```diff
@@ -0,0 +1,54 @@
...
+        thrown.expect(InvalidStateTransitionException.class);
+        thrown.expectMessage("Transition WEBHOOK_ACTION_PAID_OUT from state NEW is not valid");
+        thrown.reportMissingExceptionWithMessage("InvalidStateTransitionException expected");
+        paymentStatesGraph.getNextValidStateForEvent(NEW, WEBHOOK_ACTION_PAID_OUT);
+    }
+
+    @Test
+    public void isValidTransition_shouldReturnTrueFromWhenTransitionIsExpected() {
+        assertThat(paymentStatesGraph.isValidTransition(PaymentState.NEW, PaymentState.ENTER_DIRECT_DEBIT_DETAILS, SupportedEvent.SHOW_ENTER_DIRECT_DEBIT_DETAILS), is(true));
+    }
```
heathd: newline?

src/main/java/uk/gov/pay/directdebit/payments/model/PaymentStatesGraph.java
simad: 👍  naming is hard... :D

PR #24, opened 2017-12-04
PP-3024 Integration tests refactoring

rauligs:
- Add Test Suite to run all the integration tests. Start the application, migrate and create database template before the tests start.

- Create Rule that restore the database for every test using the created template.

- Change _DropwizarAppWithPostgres_ to Environment rule only providing configuration and port where it runs.
 
- Migrations running as part of app startup to be extracted in next commit/pr but good for now.


src/test/java/uk/gov/pay/directdebit/infra/PostgresTemplate.java
```diff
@@ -0,0 +1,76 @@
...
+        private final String databaseRootUri;
+
+        private PostgresConfig(String username, String password, String databaseName, String databaseRootUri) {
+            this.username = username;
+            this.password = password;
+            this.databaseName = databaseName;
+            this.databaseRootUri = databaseRootUri;
+        }
+
+        private static final PostgresConfig valueOf(DataSourceFactory datasource) {
```
simad: I would call this something like `from` rather than `valueOf`, which is a bit misleading. Not a deal breaker, can change in another PR.

src/test/java/uk/gov/pay/directdebit/infra/PostgresContainer.java
```diff
@@ -20,72 +20,82 @@
...
+    }
+
+    void stop() {
+        if (!stopped) {
+            try {
+                stopped = true;
+                LOG.info("Killing postgres container with ID: " + containerId);
+                LogStream logs = docker.logs(containerId, DockerClient.LogsParam.stdout(), DockerClient.LogsParam.stderr());
+                LOG.info("Killed container logs:\n");
+                logs.attach(System.out, System.err);
```
simad: I would've expected us to do a `LOG.info(logs.readFully())` or something, if we wants the logs to show up. What are we doing with the docker logs, instead?

src/test/java/uk/gov/pay/directdebit/infra/PostgresTemplate.java
rauligs: Will do

src/test/java/uk/gov/pay/directdebit/infra/PostgresContainer.java
rauligs: I'll have a look and raise a new PR for this ... need to investigate a bit

simad: (Y)

PR #25, opened 2017-12-05
PP-2971 Add Payment Request Service and Resource

simad:
- Renamed `charge_id` of tokens table to `payment_request_id`
- Added External States mapping to internal states. This mapping is just a start and will be tweaked moving forward.
- DD connector generates a new token only if the charge transaction associated to a payment request is not in a final state
- Added a payment service and a payment request resource, which serves the endpoints `/accounts/{accountId}/charges` to create a new charge, and `/accounts/{accountId}/charges/{paymentRequestId}` to get an existing payment. The reason the endpoint is still `/charges/` is for compatibility with the current implementation of public api, which selects the correct connector url and appends the path to that. 
- Still need to implement all the gateway accounts bits in dd connector


src/main/java/uk/gov/pay/directdebit/payments/api/PaymentRequestResponse.java
```diff
@@ -0,0 +1,124 @@
...
+        result = 31 * result + (returnUrl != null ? returnUrl.hashCode() : 0);
+        result = 31 * result + (description != null ? description.hashCode() : 0);
+        result = 31 * result + (reference != null ? reference.hashCode() : 0);
+        result = 31 * result + (createdDate != null ? createdDate.hashCode() : 0);
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        // Some services put PII in the description, so don’t include it in the stringification
```
rauligs: Rather than a comment I would add a test that specifically check that description field is not present when doing toString (with proper named method).

src/main/java/uk/gov/pay/directdebit/payments/resources/PaymentRequestResource.java
```diff
@@ -0,0 +1,63 @@
...
+        return paymentRequestService.getPaymentWithExternalId(paymentRequestExternalId, uriInfo)
+                .map(chargeResponse -> Response.ok(chargeResponse).build())
+                .orElseGet(() -> responseWithChargeNotFound(paymentRequestExternalId));
+    }
+
+    @POST
+    @Path(CHARGES_API_PATH)
+    @Produces(APPLICATION_JSON)
+    public Response createNewPaymentRequest(@PathParam(ACCOUNT_ID) Long accountId, Map<String, String> paymentRequest, @Context UriInfo uriInfo) {
+        //this is a good case for an Either :P
```
rauligs: This is a very good case to use exception mappers so not needed a switch, either... Resources should be a 2 liner thing but we can discuss it (see document of design principles for Dropwizard)

src/main/java/uk/gov/pay/directdebit/payments/services/PaymentRequestService.java
```diff
@@ -0,0 +1,147 @@
...
+                "type", type,
+                "params", params
+        );
+    }
+
+    public PaymentRequestResponse populateResponseWith(PaymentRequest paymentRequest, Transaction charge, UriInfo uriInfo) {
+        String paymentRequestExternalId = paymentRequest.getExternalId();
+        List<Map<String, Object>> dataLinks = new ArrayList<>();
+        dataLinks.add(createLink("self", GET, selfUriFor(uriInfo, paymentRequest.getGatewayAccountId(), paymentRequestExternalId)));
+        if (!charge.getState().toExternal().isFinished()) {
```
rauligs: So next_url is being used only for getting one single charge but not a list, if this is copied from connector we could change that for direct debit connector

src/test/java/uk/gov/pay/directdebit/payments/resources/PaymentRequestResourceTest.java
```diff
@@ -0,0 +1,180 @@
...
+                .body("links", containsLink("self", "GET", documentLocation))
+                .body("links", containsLink("next_url", "GET", newHrefNextUrl))
+                .body("links", containsLink("next_url_post", "POST", hrefNextUrlPost, "application/x-www-form-urlencoded", new HashMap<String, Object>() {{
+                    put("chargeTokenId", newChargeTokenId);
+                }}));
+
+    }
+
+    @Test
+    public void shouldReturn400IfMandatoryFieldsMissing() {
```
rauligs: With Exception mappers we wouldn't need to test this at integration level and fix issues we have in other projects having too many integration tests and few unit (when it should be the other way around) but can be done in next PR

src/test/java/uk/gov/pay/directdebit/payments/services/PaymentRequestServiceTest.java
```diff
@@ -0,0 +1,267 @@
...
+        put("return_url", RETURN_URL);
+        put("description", DESCRIPTION);
+        put("reference", REFERENCE);
+    }};
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Mock
+    private TokenDao mockedTokenDao;
```
rauligs: Too many Mock implies that this class is doing way too much. But something we can discuss to do things differently for direct debit connector

src/main/java/uk/gov/pay/directdebit/common/exception/ApiException.java
```diff
@@ -0,0 +1,19 @@
...
+
+import org.slf4j.Logger;
+import uk.gov.pay.directdebit.app.logger.PayLoggerFactory;
+
+public class ApiException extends RuntimeException {
+    private static final Logger LOGGER = PayLoggerFactory.getLogger(ApiException.class);
+    private Integer statusCode;
+
+    Integer getStatusCode() {
+        return statusCode;
```
rauligs: The status of a response does not belong to an exception but the mapper. So mapper receives an exception and decides what status code to send (an invoker of an exception shouldn't now about status codes in responses)

src/main/java/uk/gov/pay/directdebit/common/exception/ApiException.java
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    ApiException(Integer statusCode, String message) {
+        super(message);
+        this.statusCode = statusCode;
+        LOGGER.error(message);
```
rauligs: I would log the error wherever it happens. Logging is easier to read since it appears where in the class was caused rather than here (I know it saves lines of code in places but we lose the point of logging where the thing happened). Logs are logged as Whatever.class "Error message happened here"

src/main/java/uk/gov/pay/directdebit/common/exception/ApiExceptionMapper.java
```diff
@@ -0,0 +1,21 @@
...
+    /**
+     * Map an exception to a {@link Response}.
+     *
+     * @param exception the exception to map to a response.
+     * @return a response mapped from the supplied exception.
+     */
+    @Override
+    public Response toResponse(ApiException exception) {
+        ImmutableMap<String, String> entity = ImmutableMap.of("message", exception.getMessage());
+        return Response.status(exception.getStatusCode()).entity(entity).build();
```
rauligs: The exception mapper decides, depending on the exception, the status code to send back. So is better to use many exceptions named appropriately (so the stackTrace is displayed on logging and is clear what and where it happens)

src/main/java/uk/gov/pay/directdebit/common/exception/BadRequestException.java
```diff
@@ -0,0 +1,8 @@
+package uk.gov.pay.directdebit.common.exception;
+
+public class BadRequestException extends ApiException {
+
+    public BadRequestException(String message) {
+        super(400, message);
```
rauligs: Same here I would expect a BadRequestExceptionMapper knowing that when it happens it returns a 400

src/main/java/uk/gov/pay/directdebit/common/exception/InternalServerErrorException.java
```diff
@@ -0,0 +1,8 @@
+package uk.gov.pay.directdebit.common.exception;
+
+public class InternalServerErrorException extends ApiException {
+
+    public InternalServerErrorException(String message) {
+        super(500, message);
```
rauligs: Same here

src/main/java/uk/gov/pay/directdebit/common/exception/NotFoundException.java
```diff
@@ -0,0 +1,8 @@
+package uk.gov.pay.directdebit.common.exception;
+
+public class NotFoundException extends ApiException {
+
+    public NotFoundException(String message) {
+        super(404, message);
```
rauligs: Same here

src/main/java/uk/gov/pay/directdebit/common/resources/V1ApiPaths.java
```diff
@@ -3,9 +3,6 @@
 public class V1ApiPaths {
 
     public static final String ROOT_PATH = "/v1";
-    //have to be /charges unless we change public api
-    public static final String CHARGE_API_PATH = ROOT_PATH +"/api/accounts/{accountId}/charges/{paymentRequestExternalId}";
-    public static final String CHARGES_API_PATH = ROOT_PATH +"/api/accounts/{accountId}/charges";
 
     public static final String WEBHOOKS_GOCARDLESS_PATH = ROOT_PATH + "/webhooks/gocardless";
```
rauligs: Are we getting rid off constants for paths? or discussing other time?

src/main/java/uk/gov/pay/directdebit/common/util/ResponseUtil.java
```diff
@@ -1,111 +0,0 @@
...
-import org.slf4j.LoggerFactory;
-
-import javax.ws.rs.core.Response;
-import java.util.List;
-
-import static java.lang.String.format;
-import static javax.ws.rs.core.Response.*;
-import static javax.ws.rs.core.Response.Status.*;
-
-public class ResponseUtil {
```
rauligs: Definitely!!!!

src/main/java/uk/gov/pay/directdebit/common/exception/InternalServerErrorExceptionMapper.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.pay.directdebit.common.exception;
+
+import com.google.common.collect.ImmutableMap;
+
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+
+public class InternalServerErrorExceptionMapper implements ExceptionMapper<InternalServerErrorException> {
+    /**
+     * Map an exception to a {@link Response}.
```
rauligs: To remove comment (next PR)

simad:
Thanks 👍

PR #27, opened 2017-12-08
PP-2971 Add endpoint for token exchange with frontend

simad:
- Adding two endpoint for validating secure tokens between frontend and connector.
  The first endpoint validates a one-time token passed in from frontend's `next_url`.
  The token is valid if it is present in the database. In that case, connector retrieves
  the associated charge `Transaction`, updates its status through the event `TOKEN_EXCHANGED`
  and records that event in the db.
- The second endpoint deletes a one-time token.


src/main/java/uk/gov/pay/directdebit/payments/dao/mapper/TransactionMapper.java
```diff
@@ -13,12 +13,14 @@
     private static final String AMOUNT_COLUMN = "amount";
     private static final String TYPE_COLUMN = "type";
     private static final String STATE_COLUMN = "state";
+    private static final String PAYMENT_REQUEST_EXTERNAL_ID = "external_id";
```
rauligs: If using all with suffix COLUMN then this one needs to be updated (for next PR is fine)

PR #36, opened 2017-12-20
PP-2974 Direct Debit confirm

rauligs:
## WHAT

- Add relevant states and events for Sandbox confirm.
- Add Mandates database table as discussed in db design using at the moment external_id as reference.
- Add new ConflictException for when states transitions are not expected / invalid given the current state and the event.
- Pending next PR return mandate reference on confirm


src/main/resources/migrations/00006_create_table_mandates.sql
```diff
@@ -0,0 +1,14 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_table-mandates
+CREATE TABLE mandates (
+    id BIGSERIAL PRIMARY KEY,
+    payer_id BIGSERIAL NOT NULL,
+    external_id VARCHAR(30) NOT NULL,
```
simad: We might need an index on external_id, but not right now it seems, so 👍

src/main/java/uk/gov/pay/directdebit/payments/model/PaymentRequestEvent.java
```diff
@@ -75,32 +106,17 @@ public void setEventDate(ZonedDateTime eventDate) {
         CHARGE_CREATED,
         TOKEN_EXCHANGED,
         DIRECT_DEBIT_DETAILS_RECEIVED,
-        PAYER_CREATED;
+        PAYER_CREATED,
```
simad: Don't we want to add `MANDATE` as a possible Event Type? That was originally the plan, I reckon.

src/main/resources/migrations/00006_create_table_mandates.sql
rauligs: Absolutely! I'll add it now

src/main/java/uk/gov/pay/directdebit/payments/model/PaymentRequestEvent.java
rauligs: Mandate is an Event type...

simad: Not according to the gods of source code :godmode:

PR #46, opened 2018-01-23
PP-3080 "/v1/api/accounts/{accountId}/charges" endpoint validation

DanailMinchev:
## WHAT

- Finished validation for `v1/api/accounts/{accountId}/charges` and (uk.gov.pay.directdebit.payments.resources.PaymentRequestResource.createNewPaymentRequest)[https://github.com/alphagov/pay-direct-debit-connector/blob/0334d002ef44775d67c69fc037ab0dba9efcab2c/src/main/java/uk/gov/pay/directdebit/payments/resources/PaymentRequestResource.java#L48] endpoint and added tests
- Removed not-needed integration tests

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/common/validation/ApiValidation.java
+++ b/src/main/java/uk/gov/pay/directdebit/common/validation/ApiValidation.java
@@ -1,6 +1,6 @@
 package uk.gov.pay.directdebit.common.validation;
 
-
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import uk.gov.pay.directdebit.app.logger.PayLoggerFactory;
 import uk.gov.pay.directdebit.common.exception.validation.InvalidFieldsException;
@@ -16,43 +16,19 @@ import java.util.stream.Collectors;
 import static java.util.Arrays.stream;
 
 public abstract class ApiValidation {
+
     private static final Logger LOGGER = PayLoggerFactory.getLogger(ApiValidation.class);
 
     private String[] requiredFields;
-    private Map<String, Integer> maximumFieldsSize;
+    private Map<String, FieldSize> fieldSizes;
     private Map<String, Function<String, Boolean>> validators;
 
-    public ApiValidation(String[] requiredFields, Map<String, Integer> maximumFieldsSize, Map<String, Function<String, Boolean>> validators) {
+    public ApiValidation(String[] requiredFields,
+                         Map<String, FieldSize> fieldSizes,
+                         Map<String, Function<String, Boolean>> validators) {
         this.requiredFields = requiredFields;
-        this.maximumFieldsSize = maximumFieldsSize;
+        this.fieldSizes = fieldSizes;
         this.validators = validators;
-    }
-
-    private boolean isFieldSizeValid(Map<String, String> chargeRequest, String fieldName, int fieldSize) {
-        Optional<String> value = Optional.ofNullable(chargeRequest.get(fieldName));
-        return !value.isPresent() || value.get().length() <= fieldSize; //already checked that mandatory fields are already there
-    }
-
-    private List<String> checkMissingFields(Map<String, String> inputData, String[] requiredFields) {
-        return stream(requiredFields)
-                .filter(field -> !inputData.containsKey(field))
-                .collect(Collectors.toList());
-    }
-
-    private List<String> checkInvalidSizeFields(Map<String, String> inputData, Map<String, Integer> maximumFieldsSize) {
-        return maximumFieldsSize.entrySet().stream()
-                .filter(entry -> !isFieldSizeValid(inputData, entry.getKey(), entry.getValue()))
-                .map(Map.Entry::getKey)
-                .collect(Collectors.toList());
-    }
-
-    private List<String> validateParams(Map<String, String> inputData, Map<String, Function<String, Boolean>> validators) {
-        return inputData.entrySet().stream()
-                .filter(entry -> Optional.ofNullable(validators.get(entry.getKey()))
-                        .map(validator -> !validator.apply(entry.getValue()))
-                        .orElse(false))
-                .map(Map.Entry::getKey)
-                .collect(Collectors.toList());
     }
 
     public void validate(String externalId, Map<String, String> request) {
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/directdebit/common/validation/ApiValidation.java
```diff
@@ -19,18 +19,54 @@
...
+        this.minimumFieldsSize = minimumFieldsSize;
         this.maximumFieldsSize = maximumFieldsSize;
         this.validators = validators;
     }
 
-    private boolean isFieldSizeValid(Map<String, String> chargeRequest, String fieldName, int fieldSize) {
-        Optional<String> value = Optional.ofNullable(chargeRequest.get(fieldName));
-        return !value.isPresent() || value.get().length() <= fieldSize; //already checked that mandatory fields are already there
+    public void validate(String externalId, Map<String, String> request) {
+        List<String> missingFields = checkMissingFields(request, requiredFields);
```
simad: We can think about refactoring this logic (run a validation, check if empty) because this will keep growing if we add more validators. For now, I guess it's ok 👍

src/main/java/uk/gov/pay/directdebit/common/validation/ApiValidation.java
```diff
@@ -19,18 +19,54 @@
...
+    public void validate(Map<String, String> request) {
+        validate("<external id not yet created>", request);
+    }
+
+    public static boolean isNumeric(String string) {
+        return (string != null) && string.matches("[0-9]+");
+    }
+
+    public static boolean isNotNullOrEmpty(String string) {
+        return !((string == null) || string.isEmpty());
```
simad: We can use StringUtils.isBlank(`<string>`) (or the equivalent guava helpers) to make it a bit easier to read

src/main/java/uk/gov/pay/directdebit/common/validation/ApiValidation.java
```diff
@@ -39,9 +75,28 @@ private boolean isFieldSizeValid(Map<String, String> chargeRequest, String field
...
+    private boolean isMinimumFieldSizeValid(Map<String, String> inputData, String fieldName, int fieldSize) {
+        Optional<String> value = Optional.ofNullable(inputData.get(fieldName));
+        return !value.isPresent() || //already checked that mandatory fields are already there
+                value.get().length() >= fieldSize;
+    }
+
+    private List<String> checkInvalidMinimumSizeFields(Map<String, String> inputData, Map<String, Integer> maximumFieldsSize) {
         return maximumFieldsSize.entrySet().stream()
-                .filter(entry -> !isFieldSizeValid(inputData, entry.getKey(), entry.getValue()))
+                .filter(entry -> !isMinimumFieldSizeValid(inputData, entry.getKey(), entry.getValue()))
```
simad: I  think having `checkInvalidMinimumSizeFields` and `checkInvalidMaximumSizeFields` is duplicating unnecessarily quite some code. We could expand the size validation to check for both minimum and maximum (ie `value.get().length() < minLength || value.get().length() > maxLength`) instead. 
Instead of `Map<String, Integer> maximumFieldsSize` and `Map<String, Integer> minimumFieldsSize` there are different approaches - a multimap, a proper class with min and max length as fields, a bifunction which takes min and max length and returns a boolean, etc. etc.

 Feel free to discuss this approach if you don't agree :)

src/main/java/uk/gov/pay/directdebit/payers/api/CreatePayerValidator.java
```diff
@@ -1,35 +1,57 @@
...
+    public final static String SORTCODE_KEY = "sort_code";
+    public final static String ACCOUNT_NUMBER_KEY = "account_number";
+    public final static String ADDRESS_COUNTRY_KEY = "country_code";
+    public final static String ADDRESS_LINE1_KEY = "address_line1";
+    public final static String ADDRESS_CITY_KEY = "city";
+    public final static String ADDRESS_POSTCODE_KEY = "postcode";
+    public final static String EMAIL_KEY = "email";
+
+    private final static Map<String, Function<String, Boolean>> validators;
+    static {
```
simad: I don't understand why we are statically initializing everything, creating a hashmap, but then switching to unmodifiableMap. Could you explain, please? Thanks :)

DanailMinchev: @simad I just wanted to use the native approach to create immutable map and not depend on 3rd party library, as described here: https://stackoverflow.com/questions/22636575/unmodifiablemap-java-collections-vs-immutablemap-google/22636750#comment65774752_22636750
as discussed I will revert

src/main/java/uk/gov/pay/directdebit/common/validation/ApiValidation.java
DanailMinchev: Thanks @simad , great idea! I've added the class approach

src/main/java/uk/gov/pay/directdebit/common/validation/ApiValidation.java
```diff
@@ -39,13 +67,19 @@ private boolean isFieldSizeValid(Map<String, String> chargeRequest, String field
...
-    private List<String> checkInvalidSizeFields(Map<String, String> inputData, Map<String, Integer> maximumFieldsSize) {
-        return maximumFieldsSize.entrySet().stream()
+    private List<String> checkFieldSizes(Map<String, String> inputData, Map<String, FieldSize> fieldSizes) {
+        return fieldSizes.entrySet().stream()
                 .filter(entry -> !isFieldSizeValid(inputData, entry.getKey(), entry.getValue()))
                 .map(Map.Entry::getKey)
                 .collect(Collectors.toList());
     }
 
+    private boolean isFieldSizeValid(Map<String, String> inputData, String fieldName, FieldSize fieldSize) {
```
simad: Looks great! 👍

src/main/java/uk/gov/pay/directdebit/payments/api/PaymentRequestValidator.java
```diff
@@ -15,19 +17,24 @@
...
+
     private final static String[] requiredFields = {AMOUNT_KEY, DESCRIPTION_KEY, REFERENCE_KEY, RETURN_URL_KEY};
-    private final static Map<String, Integer> maximumFieldsSize = ImmutableMap.of(
-            DESCRIPTION_KEY, 255,
-            REFERENCE_KEY, 255
-    );
+
+    private final static Map<String, FieldSize> fieldSizes =
+            ImmutableMap.<String, FieldSize>builder()
+                    .put(DESCRIPTION_KEY, new FieldSize(0, 255))
```
simad: Cool! This way max and min are explicit

src/main/java/uk/gov/pay/directdebit/payments/api/PaymentRequestValidator.java
```diff
@@ -15,19 +17,24 @@
...
-            AMOUNT_KEY, (amount) -> {
-                Integer amountValue = Integer.valueOf(amount);
-                return MIN_AMOUNT <= amountValue && MAX_AMOUNT >= amountValue;
-            }
-    );
+    private final static Map<String, Function<String, Boolean>> validators =
+            ImmutableMap.<String, Function<String, Boolean>>builder()
+                    .put(AMOUNT_KEY, (amount) -> {
+                        Integer amountValue = Integer.valueOf(amount);
+                        return MIN_AMOUNT <= amountValue && MAX_AMOUNT >= amountValue;
```
simad: I guess that this technically could be a size validator (with max / min size, as you did before) but I guess we can keep it like this for now

PR #48, opened 2018-01-24
PP-3080 Change max amount validation

DanailMinchev:
## WHAT

- Change max amount to `500000` pence which is `5000` pounds
- Small refactoring of the names


src/main/java/uk/gov/pay/directdebit/payments/api/PaymentRequestValidator.java
```diff
@@ -15,7 +15,7 @@
     private final static String REFERENCE_KEY = "reference";
 
     private final static int MIN_AMOUNT = 1;
-    private final static int MAX_AMOUNT = 10000000;
+    private final static int MAX_AMOUNT = 500000;
```
alexbishop1: I’m going to be really picky here and say this would be clearer with underscores e.g. `5000_00` (yes, you can do that in Java!).

Also, if it’s not too much trouble could you rename this constant to be `MAX_AMOUNT_IN_PENCE`? It’s always nice when variable names describe their units (it would be even nicer to use a specific type but let’s not go there right now).

PR #51, opened 2018-01-24
BAU: Use govukpay openjdk image

tlwr:
We should use the openjdk image from govukpay, this will give us chamber as well as our other additions.


heathd:
👍

PR #54, opened 2018-02-01
PP-3061 Add gateway accounts services and resource

simad:
## WHAT
 - We are generating an *external id* for the gateway account, and assuming that scripts / self service / public auth will use that, rather than the internal id. This is needed because we now have gateway accounts in two places (cc and dd connectors), and using  internal ids will cause clashes in accounts, as the ids are generated sequentially from two different dbs.
We are going to raise a few other PRs in those microservices to handle external ids and internal ids at the same time.
- When we have to do any operation on a charge/payment request, we first check if the gateway account associated to that resource exists. If not, we throw a `GatewayAccountNotFoundException`
- When creating a gateway account, we are assuming `service_name`, `payment_provider` and `type` are mandatory.

 Note: The gateway account support is not complete: it needs a few extra endpoints (ie. a PATCH endpoint to change service name and so on). 
This will be addressed in another story, as it is not a blocker for creating API keys.

with @alexbishop1

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/gatewayaccounts/api/CreateGatewayAccountValidator.java
+++ b/src/main/java/uk/gov/pay/directdebit/gatewayaccounts/api/CreateGatewayAccountValidator.java
@@ -1,0 +1,36 @@
+package uk.gov.pay.directdebit.gatewayaccounts.api;
+
+import com.google.common.collect.ImmutableMap;
+import uk.gov.pay.directdebit.common.validation.ApiValidation;
+import uk.gov.pay.directdebit.common.validation.FieldSize;
+
+import java.util.Map;
+import java.util.function.Function;
+
+public class CreateGatewayAccountValidator extends ApiValidation {
+
+    private static final String PAYMENT_PROVIDER_KEY = "payment_provider";
+    private static final String SERVICE_NAME_KEY = "service_name";
+    private static final String TYPE_KEY = "type";
+
+    private static final int SERVICE_NAME_FIELD_LENGTH = 50;
+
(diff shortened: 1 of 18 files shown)


src/main/java/uk/gov/pay/directdebit/gatewayaccounts/resources/GatewayAccountResource.java
```diff
@@ -0,0 +1,73 @@
...
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.UriInfo;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+@Path("/")
```
DanailMinchev: Do we need this path at root level? If not, maybe it's good to remove them from all resources (probably in a separate clean up commit). Thanks

src/main/java/uk/gov/pay/directdebit/gatewayaccounts/api/CreateGatewayAccountValidator.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Map;
+import java.util.function.Function;
+
+public class CreateGatewayAccountValidator extends ApiValidation {
+
+    private static final String PAYMENT_PROVIDER_KEY = "payment_provider";
+    private static final String SERVICE_NAME_KEY = "service_name";
+    private static final String TYPE_KEY = "type";
+
+    private static final int SERVICE_NAME_FIELD_LENGTH = 50;
```
DanailMinchev: As discussed we should clarify the service name validation logic and possibly remove service name from direct debit connector

PR #60, opened 2018-02-20
PP-2886 Create customer and bank accounts for gocardless

simad:
## WHAT
 - Introducing gocardless as a payment provider. We are making the create customer and bank account after having created our own `Payer` entity. 
 - Introducing the a PaymentProviderMapper which returns a sandbox or a gocardless
   service, depending on the payment provider of the transaction. Sandbox does not
   make any call.
 - Wrapping the gocardless client in a thin wrapper, to protect ourselves from api changes
   and not have GoCardless-specific code in our resources. 
 - Introducing wiremock to test locally calls to GoCardless, with partially pre-canned responses
 - We are going to connect to stubs for local development by overriding the clientUrl in the GoCardless client.

Note: adding a nasty hack to inject our ssl certificates in the gocardless client. We are manipulating the underlying http client, which should be private. We asked GoCardless if there's a better way of using our trust loader with their client.


src/main/java/uk/gov/pay/directdebit/payments/services/SandboxService.java
```diff
@@ -0,0 +1,17 @@
...
+public class SandboxService implements DirectDebitPaymentProvider {
+
+
+    public SandboxService() {
+    }
+
+
+    @Override
+    public String createCustomer(String paymentRequestExternalId, Payer payer, String sortCode, String accountNumber) {
+        return null;
```
rauligs: This is an interface segregation. Is fine returning a service but this must be a top level one. From our point of view we are interested on payers and customers are a Gocardless thing so I for me I would see:
A real DirectDebitService has a method to createPayer then two impls Gocardless that actually creates a payer and a customer and Sandbox that only creates a payer. So the interface is not segregated.This means to discover the provider at top level (as soon the resource is hit)

src/main/resources/migrations/00009_create_table_gocardless_customers.sql
```diff
@@ -3,6 +3,7 @@
 --changeset uk.gov.pay:add_table-gocardless_customers
 CREATE TABLE gocardless_customers (
     id BIGSERIAL PRIMARY KEY,
+
```
DanailMinchev: new line not needed

src/main/java/uk/gov/pay/directdebit/payments/clients/GoCardlessClientWrapper.java
```diff
@@ -0,0 +1,44 @@
...
+        this.goCardlessClient = goCardlessClient;
+    }
+
+    public GoCardlessCustomer createCustomer(String paymentRequestExternalId, Payer payer) {
+        Customer customer = goCardlessClient.customers()
+                .create()
+                .withEmail(payer.getEmail())
+                .withGivenName(payer.getName())
+                .withFamilyName(payer.getName())
+                .withIdempotencyKey(paymentRequestExternalId)
```
DanailMinchev: Great that we use `paymentRequestExternalId` for idempotency key 👍

src/main/java/uk/gov/pay/directdebit/DirectDebitConnectorApp.java
```diff
@@ -93,14 +103,23 @@ public void run(DirectDebitConfig configuration, Environment environment) throws
...
         PaymentRequestEventDao paymentRequestEventDao = jdbi.onDemand(PaymentRequestEventDao.class);
         TransactionDao transactionDao = jdbi.onDemand(TransactionDao.class);
         PayerDao payerDao = jdbi.onDemand(PayerDao.class);
         MandateDao mandateDao = jdbi.onDemand(MandateDao.class);
         GatewayAccountDao gatewayAccountDao = jdbi.onDemand(GatewayAccountDao.class);
+        GoCardlessCustomerDao goCardlessCustomerDao = jdbi.onDemand(GoCardlessCustomerDao.class);
+
+        // clients
+        GoCardlessClient goCardlessClient = hackGoCardlessClient(configuration, socketFactory);
+        GoCardlessClientWrapper goCardlessClientWrapperWrapper = new GoCardlessClientWrapper(goCardlessClient);
```
rauligs: WrapperWraper 🌊

src/main/java/uk/gov/pay/directdebit/payers/resources/PayerResource.java
```diff
@@ -22,27 +25,30 @@
...
+        this.paymentProviderFactory = paymentProviderFactory;
     }
 
     @POST
-    @Path(PAYERS_API_PATH)
+    @Path("/v1/api/accounts/{accountId}/payment-requests/{paymentRequestExternalId}/payers")
     @Consumes(APPLICATION_JSON)
     @Produces(APPLICATION_JSON)
-    public Response createPayer(@PathParam("accountId") Long internalAccountId, @PathParam("paymentRequestExternalId") String paymentRequestExternalId, Map<String, String> createPayerRequest, @Context UriInfo uriInfo) {
+    public Response createPayer(@PathParam("accountId") GatewayAccount gatewayAccount, @PathParam("paymentRequestExternalId") String paymentRequestExternalId, Map<String, String> createPayerRequest, @Context UriInfo uriInfo) {
```
rauligs: _For next PR_,  I assume this is the body as map, we can convert this to actual entities or convert and validate in the resource the map to an instance of PayerRequest (or something like that)

src/main/resources/config/config.yaml
```diff
@@ -29,7 +29,8 @@ graphite:
 goCardless:
   # For initial integration we will use sandbox access token.
   # When in live mode: access token should be removed from the config and we should use partner integration instead.
-  accessToken: ${GDS_DIRECTDEBIT_CONNECTOR_GOCARDLESS_ACCESS_TOKEN:-}
+  accessToken: ${GDS_DIRECTDEBIT_CONNECTOR_GOCARDLESS_ACCESS_TOKEN:-change-me}
+  clientUrl: ${GDS_DIRECTDEBIT_CONNECTOR_GOCARDLESS_URL:-https://stubs.pymnt.localdomain/stub/gocardless}
```
rauligs: To consider for next PR, Given that this will go eventually to production environment I wouldn't default to a test value (I would set by default to a production URL and forcing the other envs to override it - this does not apply to secrets of course)

src/test/java/uk/gov/pay/directdebit/payments/resources/PaymentRequestResourceIT.java
```diff
@@ -178,10 +178,10 @@ public void shouldReturn400IfFieldsInvalidSize() {
     }
 
     @Test
-    public void shouldReturn400IfFieldsInvalid() {
+    public void shouldReturn400IfFieldsInvalid() throws JsonProcessingException {
         String accountId = testGatewayAccount.getId().toString();
 
-        String postBody = gson.toJson(ImmutableMap.builder()
+        String postBody = new ObjectMapper().writeValueAsString(ImmutableMap.builder()
```
rauligs: Minor - For next PR, ObjectMapper is an expensive object to build every time, is test code so is fine but can we added in the Before

src/test/java/uk/gov/pay/directdebit/util/WiremockStubs.java
```diff
@@ -0,0 +1,62 @@
...
+import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
+import static javax.ws.rs.core.HttpHeaders.CONTENT_TYPE;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static uk.gov.pay.directdebit.util.TestRequestResponsesLoader.GOCARDLESS_CREATE_CUSTOMER_BANK_ACCOUNT_REQUEST;
+import static uk.gov.pay.directdebit.util.TestRequestResponsesLoader.GOCARDLESS_CREATE_CUSTOMER_BANK_ACCOUNT_SUCCESS_RESPONSE;
+import static uk.gov.pay.directdebit.util.TestRequestResponsesLoader.GOCARDLESS_CREATE_CUSTOMER_REQUEST;
+import static uk.gov.pay.directdebit.util.TestRequestResponsesLoader.GOCARDLESS_CREATE_CUSTOMER_SUCCESS_RESPONSE;
+import static uk.gov.pay.directdebit.util.TestRequestResponsesLoader.load;
+
+public class WiremockStubs {
```
rauligs: For next PR, I wouldn't mix class names with specific technology, Wiremock  and Stubs are the same thing, looks like this is more like _GocardlessStubs_

src/main/resources/config/config.yaml
simad: yes, in the pr i am about to release I am removing the default for that, if it's not set, it will use the gocardless url for the account (TEST / LIVE)

src/test/java/uk/gov/pay/directdebit/payments/resources/PaymentRequestResourceIT.java
simad: sure!

src/test/java/uk/gov/pay/directdebit/util/WiremockStubs.java
simad: 👍

src/main/java/uk/gov/pay/directdebit/payers/resources/PayerResource.java
simad: yes, planning to do something about that. thanks!

PR #62, opened 2018-02-26
PP-2890 Add flow for creating mandate and payment 

simad:
## WHAT
- After a customer and a customer bank account are created, and the user confirms the payment in frontend, we create a mandate with gocardless, followed by a payment. Both of these calls are asynchronous, this means we receive details about their status after days. Receiving webhooks is part of a different story.
- Overall, we have separate entities for our internal model (eg. `Payer`, `Mandate`, `Transaction`) and gocardless-specific ones, (eg. `Customer`, `CustomerBankAccount`). We also have "link objects" between our model and gocardless, (eg. `GoCardlessMandate`, `GoCardlessPayment`) which link our internal ids to the ones provided by GoCardless.
- We apply the 'hack' to inject a custom ssl trust store only when calling stubs, as this impacts talking to the real gocardless.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/model/GoCardlessMandate.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/model/GoCardlessMandate.java
@@ -1,0 +1,59 @@
+package uk.gov.pay.directdebit.mandate.model;
+
+import com.google.common.base.Objects;
+
+public class GoCardlessMandate {
+
+    private Long id;
+    private Long mandateId;
+    private String goCardlessMandateId;
+
+
+    public GoCardlessMandate(Long id, Long mandateId, String goCardlessMandateId) {
+        this.id = id;
+        this.mandateId = mandateId;
+        this.goCardlessMandateId = goCardlessMandateId;
+    }
+    public GoCardlessMandate(Long mandateId, String goCardlessMandateId) {
+        this(null, mandateId, goCardlessMandateId);
+    }
+
+    public Long getId() {
+        return id;
+    }
+
+    public void setId(Long id) {
+        this.id = id;
+    }
+
+    public Long getMandateId() {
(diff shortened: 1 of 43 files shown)


src/main/java/uk/gov/pay/directdebit/mandate/model/GoCardlessMandate.java
```diff
@@ -0,0 +1,59 @@
...
+    public void setGoCardlessMandateId(String goCardlessMandateId) {
+        this.goCardlessMandateId = goCardlessMandateId;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        GoCardlessMandate that = (GoCardlessMandate) o;
+        return Objects.equal(id, that.id) &&
```
rauligs: Why no the generated primitive one? I have had bad experiences with other ways implementing equals / hashcode with external dependencies (at that time was someone using Apache commons reflection equals). _Intellij_ even does it for us (this applies for any other equals / hashcode implemented using external dependencies). Happy to discuss / approve this and do the work in a next PR if agreed

src/main/java/uk/gov/pay/directdebit/mandate/services/PaymentConfirmService.java
```diff
@@ -4,13 +4,40 @@
...
+import uk.gov.pay.directdebit.payers.model.Payer;
 import uk.gov.pay.directdebit.payments.model.Transaction;
 import uk.gov.pay.directdebit.payments.services.TransactionService;
 
 import java.util.Optional;
 
 public class PaymentConfirmService {
 
+    public static class ConfirmationDetails {
+        private Transaction transaction;
```
rauligs: It looks like It could be a class on its own

PR #66, opened 2018-03-02
Bc pp 3443 [Do not merge]

belindac:
Java to prove container can talk to GC through the proxy and work out why it isn't working in demoservice


belindac:
Standalone java class to test GoCardlessClient, not required.

PR #77, opened 2018-03-16
PP-3238 Add happy path events handling

rauligs:
## WHAT
Handle events:
- Mandates: 'created', 'submitted','active'
- Payments: 'created', 'submitted', 'confirmed', 'paid_out'
- Transitions are pending, this is because we need to think if for one off we should add mandate transitions at all or just store the events and only _listen_ to payment ones.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
+++ b/src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
@@ -14,14 +14,18 @@ import javax.inject.Inject;
 import java.util.List;
 
 public class WebhookGoCardlessService {
+
     private static final Logger LOGGER = PayLoggerFactory.getLogger(WebhookGoCardlessService.class);
-    private final TransactionService transactionService;
+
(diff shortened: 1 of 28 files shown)


src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
```diff
@@ -14,14 +14,18 @@
...
     private final GoCardlessService goCardlessService;
+    private final GoCardlessPaymentHandler goCardlessPaymentHandler;
+    private final GoCardlessMandateHandler goCardlessMandateHandler;
 
     @Inject
     public WebhookGoCardlessService(GoCardlessService goCardlessService, TransactionService transactionService) {
         this.goCardlessService = goCardlessService;
-        this.transactionService = transactionService;
+        goCardlessPaymentHandler = new GoCardlessPaymentHandler(transactionService, goCardlessService);
+        goCardlessMandateHandler = new GoCardlessMandateHandler(transactionService, goCardlessService);
```
simad: I need to pass in the event to the handler in my PRs, so I will probably move this back to the switch statement, but it's all right for now :)

PR #78, opened 2018-03-16
PP-3443 Run a payment flow with GoCardless Sandbox

belindac:
Remove the cacerts testing as we are going to put the cacert required
into the cacert file when the instance is built, see
https://github.com/alphagov/pay-infra/pull/1396/files

solo @belindac


belindac:
I tested on my environment with the standalone client jar not specifying any cacerts and the list of customers is displayed confirming we are talking successfully

PR #79, opened 2018-03-20
PP-3360 Handle mandate failed webhook and send email to user

simad:
## WHAT
- Added the `UserNotificationService` which, similarly to connector, connects to Notify to send out emails to the user, and relative configs
- Handling `mandate failed` event. When we receive a mandate failed, it means the mandate was never actually in an active state, so GC wouldn't have sent out any payment to the bank.
This means we register two events, `MANDATE_FAILED` and `PAYMENT_FAILED`. The latter updates the transaction status to `FAILED`.


src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private void readMandateFailedEmailConfig(NotifyClientFactory notifyClientFactory) {
+        emailNotifyGloballyEnabled = notifyClientFactory.isEmailNotifyEnabled();
+        mandateFailedTemplateId = notifyClientFactory.getMandateFailedTemplateId();
+
+        if (!emailNotifyGloballyEnabled) {
+            LOGGER.warn("Email notifications are globally disabled by configuration");
+        }
+        if (emailNotifyGloballyEnabled && isBlank(mandateFailedTemplateId)) {
```
rauligs: Force the check in the configuration itself rather than here.

src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
```diff
@@ -0,0 +1,104 @@
...
+    private final GatewayAccountService gatewayAccountService;
+
+    @Inject
+    public UserNotificationService(DirectDebitConfig directDebitConfig,
+                                   NotificationClient notificationClient,
+                                   MetricRegistry metricRegistry,
+                                   GatewayAccountService gatewayAccountService) {
+        readMandateFailedEmailConfig(directDebitConfig.getNotifyConfig());
+        if (emailNotifyGloballyEnabled) {
+            int numberOfThreads = directDebitConfig.getExecutorServiceConfig().getThreadsPerCpu() * getRuntime().availableProcessors();
```
rauligs: Something that can be extracted in the configuration mapping class

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/GoCardlessMandateHandler.java
```diff
@@ -1,56 +1,71 @@
...
+        return Optional.ofNullable(GoCardlessMandateAction.fromString(event.getAction()))
+                .map((action) -> getHandledActions().get(action))
+                .map((handledAction -> {
+                    GoCardlessMandate goCardlessMandate = goCardlessService.findMandateForEvent(event);
+                    Transaction transaction = transactionService.findTransactionForMandateId(goCardlessMandate.getMandateId());
+                    return handledAction.apply(transaction);
+                }));
+    }
+
+    private PaymentRequestEvent upsertMandatePendingEvent(Transaction transaction) {
```
rauligs: In reality is not an upsert (update or insert), is more like an ignorsert! 🔨

src/main/resources/config/config.yaml
```diff
@@ -34,13 +34,23 @@ goCardless:
   webhookSecret: ${GDS_DIRECTDEBIT_CONNECTOR_GOCARDLESS_WEBHOOK_SECRET:-change-me}
   environment: ${GDS_DIRECTDEBIT_CONNECTOR_GOCARDLESS_ENVIRONMENT:-sandbox}
 
+notify:
+  apiKey: ${NOTIFY_API_KEY:-api_key-pay-notify-service-id-pay-notify-secret-needs-to-be-32-chars-fsghdngfhmhfkrgsfs}
```
rauligs: I would expect a pay-scripts  and variables to be set for the environment (just checking)

src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
simad: It's a bit tricky because of the runtime stuff, will take a look in the next PR

src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
simad: done 👍

PR #81, opened 2018-03-21
Bc pp 3443 and 3265

belindac:
Testing out sneak preview GoCardless client with the ability to use a proxy and a custom ssl socket factory that we had to work around
Use new separate split out jobs for deployment, smoke tests, tagging and
deployment event notification
Updated deployEcs arguments not to run tests and tag


belindac:
Closing as we will do this testing under a new ticket

PR #90, opened 2018-03-23
PP-2810 Add charge date column to gocardless payments table

simad:
## WHAT
This is sent to us from gocardless and it's going to be used in the body of the payment confirmation email we send to the user.

### DIFF
--- a/src/main/resources/migrations/00014_alter_table_gocardless_payments_charge_date.sql
+++ b/src/main/resources/migrations/00014_alter_table_gocardless_payments_charge_date.sql
@@ -1,0 +1,5 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:alter_table-gocardless-payments-charge-date
+ALTER TABLE gocardless_payments ADD COLUMN charge_date DATE NOT NULL DEFAULT NOW();
+--rollback ALTER TABLE gocardless_payments DROP COLUMN charge_date;


src/main/resources/migrations/00014_alter_table_gocardless_payments_charge_date.sql
```diff
@@ -0,0 +1,5 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:alter_table-gocardless-payments-charge-date
+ALTER TABLE gocardless_payments ADD COLUMN charge_date TIMESTAMP;
```
alexbishop1: Do we usually use `TIMESTAMP` for things like this?

simad: We usually use `TIMESTAMP WITH TIME ZONE DEFAULT (now() AT TIME ZONE 'utc')`, but in this case gocardless' date is of format "2014-05-21", so no timezone there

alexbishop1: So do we want `DATE` then?

simad: That's why I was thinking of using LocalDate in the code to represent it, instead of ZonedDateTime

simad: Oh cool, I just discovered that `TIMESTAMP` has resolution of 1 microsecond and `DATE` of 1 day. Considering that GC just gives back the day, I guess `DATE` fits this case better :) will change, thanks

alexbishop1: Definitely use `LocalDate` in the code: precise types are awesome in all kinds of ways!

simad: 👍

PR #97, opened 2018-03-29
Fix payout paid webhook not parsed properly

simad:
## WHAT
 - There was no `PAYOUTS` case when extracting the resource id, so the
   webhook for payout paid was not handled properly

## HOW 
_Steps to test or reproduce:_


src/test/java/uk/gov/pay/directdebit/webhook/gocardless/api/GoCardlessWebhookParserTest.java
```diff
@@ -18,6 +18,7 @@
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.hamcrest.core.Is.is;
 import static org.junit.Assert.assertThat;
+import static uk.gov.pay.directdebit.payments.model.GoCardlessResourceType.*;
```
alexbishop1: Static imports are the source of all evil in this world!

PR #104, opened 2018-04-12
PP-3545 Use gateway account external id

DanailMinchev:
## WHAT

This PR exposes external gateway account id which will be needed to query `/v1/api/accounts/{accountId}` endpoint.
It also changes the JDBI queries to be explicit.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/payments/dao/mapper/TransactionMapper.java
+++ b/src/main/java/uk/gov/pay/directdebit/payments/dao/mapper/TransactionMapper.java
@@ -10,22 +10,23 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 
 public class TransactionMapper implements ResultSetMapper<Transaction> {
-    private static final String ID_COLUMN = "id";
-    private static final String AMOUNT_COLUMN = "amount";
-    private static final String TYPE_COLUMN = "type";
-    private static final String STATE_COLUMN = "state";
+
+    private static final String TRANSACTION_ID_COLUMN = "transaction_id";
+    private static final String TRANSACTION_AMOUNT_COLUMN = "transaction_amount";
+    private static final String TRANSACTION_TYPE_COLUMN = "transaction_type";
+    private static final String TRANSACTION_STATE_COLUMN = "transaction_state";
     private static final String PAYMENT_REQUEST_ID_COLUMN = "payment_request_id";
-    private static final String PAYMENT_REQUEST_EXTERNAL_ID_COLUMN = "external_id";
-    private static final String PAYMENT_REQUEST_REFERENCE_COLUMN = "reference";
-    private static final String PAYMENT_REQUEST_RETURN_URL_COLUMN = "return_url";
-    private static final String PAYMENT_REQUEST_GATEWAY_ACCOUNT_ID_COLUMN = "gateway_account_id";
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/pay/directdebit/payments/dao/mapper/TransactionMapper.java
```diff
@@ -10,31 +10,35 @@
 import java.sql.SQLException;
 
 public class TransactionMapper implements ResultSetMapper<Transaction> {
-    private static final String ID_COLUMN = "id";
-    private static final String AMOUNT_COLUMN = "amount";
-    private static final String TYPE_COLUMN = "type";
-    private static final String STATE_COLUMN = "state";
+
+    private static final String TRANSACTION_ID_COLUMN = "transaction_id";
+    private static final String TRANSACTION_AMOUNT_COLUMN = "transaction_amount";
```
simad: Good job! I think I actually had a todo somewhere saying that we were overriding transaction and payment request amounts because we were not explicit as to which one we are picking 👍

simad: Actually, the todo was `//todo refactor transaction to contain a PaymentRequest`. One improvement we can make to the code is to pass a payment request object directly in Transaction, instead of having various payment requests fields (ie. external id, etc.) there. But we can do it in another pr, no probs

src/main/java/uk/gov/pay/directdebit/payments/model/Transaction.java
```diff
@@ -15,17 +15,30 @@
...
+    private String gatewayAccountExternalId;
     private PaymentProvider paymentProvider;
     private Long amount;
     private Type type;
     private PaymentState state;
 
-    public Transaction(Long id, Long paymentRequestId, String paymentRequestExternalId, String paymentRequestDescription, String paymentRequestReference, Long paymentRequestGatewayAccountId, PaymentProvider paymentProvider, String paymentRequestReturnUrl, Long amount, Type type, PaymentState state) {
+    public Transaction(Long id,
+                       Long paymentRequestId,
+                       String paymentRequestExternalId,
```
simad: This is what I am talking about. I see Transaction having an entire PaymentRequest object and potentially a GatewayAccount in its constructor. But again, it's ok to leave it as it is, for now.

src/test/java/uk/gov/pay/directdebit/payments/fixtures/TransactionFixture.java
```diff
@@ -10,16 +10,29 @@
 import uk.gov.pay.directdebit.payments.model.Transaction;
 
 public class TransactionFixture implements DbFixture<TransactionFixture, Transaction> {
+
     private Long id = RandomUtils.nextLong(1, 99999);
+
     private Long paymentRequestId = RandomUtils.nextLong(1, 99999);
+
     private String paymentRequestExternalId = RandomIdGenerator.newId();
+
```
simad: yo dawg I heard you like newlines

src/main/java/uk/gov/pay/directdebit/payments/dao/mapper/TransactionMapper.java
DanailMinchev: Yes, will do the refactoring after selfservice integration, thanks

PR #109, opened 2018-04-13
PP-3361 Handle mandate cancelled event

simad:
## WHAT
- Registering a specific event when a payment is submitted by GC to the bank

- When we receive a mandate cancelled from GoCardless, we will act depending on two different scenarios:

    - The payment associated to the mandate is CREATED, this means it has not been submitted to the bank by GC. This also means they won't send us a webhook to tell us that the payment failed, so we have to update the charge status ourselves
    - The payment associated to the mandate is SUBMITTED. We don't update the charge status ourselves as we will receive a webhook from GC telling us what happened to it.
Right now we have no concept of a lifecycle for mandates, like we do for payments. I will raise another PR to add that graph. Once that is done, mandate cancelled will update the status of a mandate.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
@@ -1,11 +1,16 @@
 package uk.gov.pay.directdebit.mandate.dao;
 
+import org.skife.jdbi.v2.sqlobject.Bind;
 import org.skife.jdbi.v2.sqlobject.BindBean;
 import org.skife.jdbi.v2.sqlobject.GetGeneratedKeys;
+import org.skife.jdbi.v2.sqlobject.SqlQuery;
 import org.skife.jdbi.v2.sqlobject.SqlUpdate;
 import org.skife.jdbi.v2.sqlobject.customizers.RegisterMapper;
+import org.skife.jdbi.v2.sqlobject.customizers.SingleValueResult;
 import uk.gov.pay.directdebit.mandate.dao.mapper.MandateMapper;
 import uk.gov.pay.directdebit.mandate.model.Mandate;
+
+import java.util.Optional;
 
 @RegisterMapper(MandateMapper.class)
 public interface MandateDao {
@@ -13,4 +18,8 @@ public interface MandateDao {
     @SqlUpdate("INSERT INTO mandates(payer_id, external_id) VALUES (:payerId, :externalId)")
     @GetGeneratedKeys
     Long insert(@BindBean Mandate mandate);
+
+    @SqlQuery("SELECT * FROM mandates m JOIN payers p ON m.payer_id = p.id JOIN transactions t ON t.payment_request_id = p.payment_request_id WHERE t.id = :transactionId")
+    @SingleValueResult(Mandate.class)
+    Optional<Mandate> findByTransactionId(@Bind("transactionId") Long transactionId);
 }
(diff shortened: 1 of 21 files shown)


src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
```diff
@@ -102,18 +104,27 @@ public UserNotificationService(DirectDebitConfig directDebitConfig,
         String mandateFailedTemplateId = directDebitConfig.getNotifyConfig().getMandateFailedTemplateId();
         LOGGER.info("Sending mandate failed email, payment request id: {}, gateway account id: {}",
                 transaction.getPaymentRequestExternalId(),
-                transaction.getGatewayAccountId());
+                transaction.getGatewayAccountExternalId());
```
DanailMinchev: 👍

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
```diff
@@ -1,16 +1,25 @@
...
+import java.util.Optional;
+
 @RegisterMapper(MandateMapper.class)
 public interface MandateDao {
 
     @SqlUpdate("INSERT INTO mandates(payer_id, external_id) VALUES (:payerId, :externalId)")
     @GetGeneratedKeys
     Long insert(@BindBean Mandate mandate);
+
+    @SqlQuery("SELECT * FROM mandates m JOIN payers p ON m.payer_id = p.id JOIN transactions t ON t.payment_request_id = p.payment_request_id WHERE t.id = :transactionId")
```
DanailMinchev: Sometimes it's a bit hard to understand which columns are mapped to which fields when we use `*` with multiple joins, for example I needed to refactor https://github.com/alphagov/pay-direct-debit-connector/pull/104/commits/ee8b4e969649a2cb922eac588672a8c504daf002 to map it correctly.
Is it good idea to do it here as well?

simad: I think it's necessary to be explicit when two or more tables have similarly named columns (ie. amount is in transaction and in payment request). You can get away with * when columns are different. I reason about that by looking at the mapper usually. If you think we should refactor I'll do it in the next pr

src/test/java/uk/gov/pay/directdebit/notifications/services/UserNotificationServiceTest.java
```diff
@@ -111,6 +115,31 @@ public void shouldSendMandateFailedEmailIfEmailNotifyIsEnabled() throws Exceptio
...
+    @Test
+    public void shouldSendMandateCancelledEmailIfEmailNotifyIsEnabled() throws Exception {
+        when(mockNotifyClientFactory.isEmailNotifyEnabled()).thenReturn(true);
+        userNotificationService = new UserNotificationService(mockDirectDebitConfig, mockNotifyClient, mockMetricRegistry, mockGatewayAccountService);
+
+        Mandate mandate = MandateFixture.aMandateFixture().toEntity();
+        HashMap<String, String> emailPersonalisation = new HashMap<>();
+        emailPersonalisation.put("mandate reference", mandate.getExternalId());
+        emailPersonalisation.put("org name", gatewayAccount.getServiceName());
+        emailPersonalisation.put("org phone", "+44 000-CAKE-000");
```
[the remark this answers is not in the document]
simad: It is not a valid number on purpose - it's a placeholder, because we need to get adminusers stuff in emails. Same for the address and so on. Is there a reason I should use a valid number (I can't think of any)?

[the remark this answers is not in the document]
DanailMinchev: I was thinking about example phone numbers, but `+44 000-CAKE-000` - `+44 000-2253-000` will not resolve I think anyway :)

PR #115, opened 2018-04-18
PP-3649 Add user cancel transaction endpoint

simad:
## WHAT
 - Adding an endpoint to cancel a transaction from frontend. This is allowed
   only if the user hasn't confirmed the payment (after which, we create mandate and payment in GC)
   So, the charge must be in:
      - `AWAITING_DIRECT_DEBIT_DETAILS` (when the user is on the insert direct
   debit details page)
      -  `PROCESSING_DIRECT_DEBIT_DETAILS` (when the user has inserted details and we are about to
   create a payer)
      -  `AWAITING_CONFIRMATION` (when the user has clicked continue, and is in the confirmation page)
-  The cancel operation simply looks for a transaction and transition its state to `CANCELLED`. We don't
   have to do any additional operation as we haven't created any charge in GC.
   This means we are going to potentially have a 'stale' payer in our DB, which is all right for now


src/main/java/uk/gov/pay/directdebit/mandate/services/PaymentConfirmService.java
```diff
@@ -31,8 +31,8 @@ public PaymentConfirmService(TransactionService transactionService, PayerDao pay
      *
      * @param paymentExternalId
      */
-    public ConfirmationDetails confirm(Long accountId, String paymentExternalId) {
-        Transaction transaction = transactionService.confirmedDirectDebitDetailsFor(accountId, paymentExternalId);
+    public ConfirmationDetails confirm(String accountExternalId, String paymentExternalId) {
```
DanailMinchev: 👍

src/main/java/uk/gov/pay/directdebit/payments/resources/PaymentRequestResource.java
```diff
@@ -50,7 +50,17 @@ public Response getCharge(@PathParam("accountId") String accountExternalId, @Pat
...
         paymentRequestValidator.validate(paymentRequest);
         LOGGER.info("Creating new payment request - {}", paymentRequest.toString());
-        PaymentRequestResponse response = paymentRequestService.createCharge(paymentRequest, accountExternalId, uriInfo);
+        PaymentRequestResponse response = paymentRequestService.createTransaction(paymentRequest, accountExternalId, uriInfo);
         return created(response.getLink("self")).entity(response).build();
     }
+
+
+    @POST
+    @Path(CANCEL_CHARGE_API_PATH)
```
DanailMinchev: May be we can use `DELETE /v1/api/accounts/{accountId}/payment-requests/{paymentRequestExternalId}` instead of `POST /v1/api/accounts/{accountId}/payment-requests/{paymentRequestExternalId}/cancel` like more restful style? On the other hand if we have `DELETED` status in the future, probably `POST` is fine as it's changing the state, not sure what is better?

PR #138, opened 2018-05-02
PP 3646 add simple search payments view endpoint

SandorArpa:
## WHAT

Implement simple search of PaymentView. Paginated search can be done by gateway account external id.

## HOW

- introduce PaymentViewDao
- add PaymentViewMapper to map search result to POJOs
- add PayersFixture
- introduce PaymentViewService
- add PaymentViewResponse
- add get methods to PayersFixture
- add equals and hashCode to PaymentView
- add PaymentViewResource to process API endpoint
- add ResponseUtil to simplify responses
- add specific validator PaymentViewValidator
- introduce PaymentViewSearchParams
- add relevant tests
- add configuration to DirectDebitModule


src/main/java/uk/gov/pay/directdebit/app/config/DirectDebitModule.java
```diff
@@ -153,4 +154,9 @@ public GoCardlessEventDao provideGoCardlessEventDao() {
         return jdbi.onDemand(GoCardlessEventDao.class);
     }
 
+    @Provides
+    @Singleton
+    public PaymentViewDao providePaymentViewDao() {
+        return new PaymentViewDao(jdbi);
```
simad: this should be `jdbi.onDemand(PaymentViewDao.class)` as all the other examples - ondemand works well with dropwizard as it opens and closes the connection to the db automatically. If we decide not to use ondemand (because of transactions and so on) and attach it to an handle, it's something we can think of later when we play the data integrity story

src/main/java/uk/gov/pay/directdebit/payments/api/PaymentViewValidator.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.stream.Collectors;
+
+import static com.google.common.collect.Lists.newArrayList;
+
+public class PaymentViewValidator {
+
+    private static final Long MAX_PAGE_NUMBER = 500l;
+    private static final Long DEFAULT_PAGE_NUMBER = 1l;
+
+    public static Pair<Long, Long> validatePagination(Optional<Pair<Long, Long>> paginationParams) {
```
simad: Not a big fan of using Optional in non public facing apis. We can just use paginationParams and check that it is not null.

simad: Also, i'd rather have a typed `PaginationParams` class, with two long fields (`maxPageNumber` and `defaultPageNumber`) rather than`Pair<Long, Long>`.

`paginationParams.getMaxPageNumber()` is really so much more clear than `paginationParams.getLeft()`

src/main/java/uk/gov/pay/directdebit/payments/api/PaymentViewValidator.java
```diff
@@ -0,0 +1,52 @@
...
+        Long pageNumber = DEFAULT_PAGE_NUMBER;
+        Long displaySize = MAX_PAGE_NUMBER;
+        if (paginationParams.isPresent()) {
+            displaySize = paginationParams.get().getRight() > MAX_PAGE_NUMBER ? MAX_PAGE_NUMBER : paginationParams.get().getRight();
+            pageNumber = (paginationParams.get().getLeft() - 1) * displaySize;
+        }
+        return Pair.of(pageNumber, displaySize);
+    }
+
+    public Optional<List> validateQueryParams(List<Pair<String, Long>> nonNegativePairMap) {
```
simad: two problems here: `List` is not typed (ie. `List<String>`) and more generally `Optional<List>` does not make much sense. A list can be empty, and that's semantically equivalent to `Optional.empty()` as we use it, no need to add so much complexity. 
So, downstream, all you have to check is that the list is empty to make sure there are no errors.

src/main/java/uk/gov/pay/directdebit/payments/api/PaymentViewValidator.java
```diff
@@ -0,0 +1,52 @@
...
+                    .stream()
+                    .map(param -> String.format(invalidQueryParams.get(param), param))
+                    .collect(Collectors.toList()));
+            return Optional.of(invalidResponse);
+        }
+        return Optional.empty();
+    }
+
+    private Map<String, String> validateNonNegativeQueryParams(List<Pair<String, Long>> nonNegativePairMap, Map<String, String> invalidQueryParams) {
+        nonNegativePairMap.forEach(param -> {
```
simad: Again, I would recommend not using `Pair` but creating a class for whatever you're representing.

src/main/java/uk/gov/pay/directdebit/payments/dao/PaymentViewDao.java
```diff
@@ -0,0 +1,46 @@
...
+import uk.gov.pay.directdebit.payments.model.PaymentView;
+
+import javax.inject.Inject;
+import java.util.List;
+
+public class PaymentViewDao {
+
+    private final Jdbi jdbi;
+
+    private StringBuilder sb = new StringBuilder(
```
simad: You don't need a string builder, this can just be a plain string

src/main/java/uk/gov/pay/directdebit/payments/dao/PaymentViewDao.java
```diff
@@ -0,0 +1,46 @@
...
+            "tr.state AS state " +
+            "FROM payment_requests pr " +
+            "JOIN payers pa ON pr.id = pa.payment_request_id " +
+            "JOIN transactions tr ON pr.id = tr.payment_request_id " +
+            "JOIN gateway_accounts ga ON ga.id = pr.gateway_account_id " +
+            "WHERE ga.external_id = :gatewayAccountExternalId " +
+            "ORDER BY pr.id DESC OFFSET :offset LIMIT :limit");
+
+    @Inject
+    public PaymentViewDao(Jdbi jdbi) {
```
simad: We have been writing dao using jdbi's fluent api (example https://github.com/alphagov/pay-direct-debit-connector/blob/master/src/main/java/uk/gov/pay/directdebit/tokens/dao/TokenDao.java) or any other. It would be good to keep using the same pattern so that we have consistency

src/main/java/uk/gov/pay/directdebit/payments/dao/PaymentViewSearchParams.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public Long getPage() {
+        return page;
+    }
+
+    public Long getDisplaySize() {
+        return displaySize;
+    }
+
+    public Optional<Pair<Long, Long>> getPaginationParams() {
```
simad: See discussions above regarding pair and optional. This can just be: `public PaginationParams getPaginationParams()`. Then, `paginationParams.getMaxSize()` or whatever, can be null and will be null checked wherever used

src/main/java/uk/gov/pay/directdebit/payments/resources/PaymentViewResource.java
```diff
@@ -0,0 +1,78 @@
...
+            @QueryParam(PAGE) Long pageNumber,
+            @QueryParam(DISPLAY_SIZE) Long displaySize) {
+        List<Pair<String, Long>> nonNegativePairMap = ImmutableList.of(Pair.of(PAGE, pageNumber), Pair.of(DISPLAY_SIZE, displaySize));
+
+        PaymentViewSearchParams searchParams = new PaymentViewSearchParams(
+                pageNumber,
+                displaySize
+        );
+        return paymentViewValidator.validateQueryParams(nonNegativePairMap)
+                .map(ResponseUtil::badRequestResponse)
```
simad: We don't need `ResponseUtil` as we've been using Dropwizard's exception mapper (see for example https://github.com/alphagov/pay-direct-debit-connector/blob/9e6eada0894167a1088655021b4bae84fc50bbd9/src/main/java/uk/gov/pay/directdebit/gatewayaccounts/exception/InvalidPaymentProviderException.java) 
All you need to to is to throw an exception that extends bad request exception in validateQueryParams, and dropwizard will intercept it and do the right thing automagically

src/main/java/uk/gov/pay/directdebit/payments/resources/PaymentViewResource.java
```diff
@@ -0,0 +1,78 @@
...
+                .orElseGet(() -> gatewayAccountDao.findByExternalId(accountExternalId)
+                        .map(gatewayAccount ->
+                                doSearch(accountExternalId, searchParams))
+                        .orElseGet(() -> notFoundResponse(format("account with id %s not found", accountExternalId)))
```
simad: I'd move all this logic in a service, to keep the resource as clean as possible. See what we've done for other resources

src/main/java/uk/gov/pay/directdebit/utils/ResponseUtil.java
```diff
@@ -0,0 +1,34 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.ws.rs.core.Response;
+import java.util.List;
+
+import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
+import static javax.ws.rs.core.Response.Status.NOT_FOUND;
+import static javax.ws.rs.core.Response.status;
+
+public class ResponseUtil {
```
simad: this class should be removed, we are using different patterns in dd connector. BadRequestResponse and NotFoundResponse should use the exception mapper and there's no need to have helpers to serialise an entity

src/test/java/uk/gov/pay/directdebit/payments/fixtures/PayersFixture.java
```diff
@@ -0,0 +1,135 @@
+package uk.gov.pay.directdebit.payments.fixtures;
```
simad: There is already https://github.com/alphagov/pay-direct-debit-connector/blob/master/src/test/java/uk/gov/pay/directdebit/payers/fixtures/PayerFixture.java, is this a duplicate?

src/main/java/uk/gov/pay/directdebit/app/config/DirectDebitModule.java
SandorArpa: I beg to differ. Declarative API is working with declared arguments. Leaves us with no flexibility to use a query based on any variation, for example email, from date, to date, state, reference, etc.

src/main/java/uk/gov/pay/directdebit/payments/api/PaymentViewValidator.java
SandorArpa: The validation is not fully done. It will be extended when I introduce to and from dates. See connector.

src/main/java/uk/gov/pay/directdebit/payments/dao/PaymentViewDao.java
SandorArpa: Yes, I'm aware that the compiler turns this String = +String into a StringBuilder under the hood. What I've done looks better and confirms to clean coding principles.

src/main/java/uk/gov/pay/directdebit/payments/dao/PaymentViewDao.java
SandorArpa: See above. We've been using declarative API. Here we need fluent API as the arguments are not static, like in the case of TokenDao.

src/main/java/uk/gov/pay/directdebit/utils/ResponseUtil.java
SandorArpa: 👍

src/test/java/uk/gov/pay/directdebit/payments/services/PaymentViewServiceTest.java
```diff
@@ -51,17 +60,39 @@ public void setUp() {
                     PaymentState.PROCESSING_DIRECT_DEBIT_DETAILS);
             paymentViewList.add(paymentView);
         }
-        paymentViewService = new PaymentViewService(paymentViewDao);
-        when(paymentViewDao.searchPaymentView(gatewayAccountExternalId, 0l, 100l)).thenReturn(paymentViewList);
+        paymentViewService = new PaymentViewService(paymentViewDao, gatewayAccountDao);
+        gatewayAccount = new GatewayAccount(1l, gatewayAccountExternalId, null, null, null, null, null);
```
simad: We normally use fixtures for this, but it's all right, we can refactor in another pr

PR #149, opened 2018-05-16
BAU: small refactorings

heathd:
see individual commits

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
+++ b/src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
@@ -14,7 +14,6 @@ public class PaymentViewSearchParams {
     private static final String TO_DATE_FIELD = "toDate";
     private static final String EMAIL_FIELD = "email";
     private static final String REFERENCE_FIELD = "reference";
-    private static final String LIKE_HOLDER = "%LIKE%";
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
```diff
@@ -107,15 +106,19 @@ public String generateQuery() {
...
+                queryMap.put(REFERENCE_FIELD, likeClause(reference));
             }
             if (amount != null) {
                 queryMap.put(AMOUNT_FIELD, amount);
             }
         }
         return queryMap;
     }
+
+    private static String likeClause(String rawUserInputText) {
```
simad: are we planning to use it outside of this class? otherwise it shouldn't be static

alexbishop1: Sometimes I like to use `static` to indicate that a method is pure and won’t change anything in this instance of the class (I know `static` in Java doesn’t always guarantee this)

heathd: IntelliJ is actually suggesting that I add 

```
    @org.jetbrains.annotations.NotNull
    @org.jetbrains.annotations.Contract(pure = true)
```

maybe we should start using these annotations?

PR #153, opened 2018-05-18
PP-3713 Update direct debit email variables

simad:
## WHAT
 - Had to refactor the code a bit because we switched payment reference to mandate reference.
   Everything else is just about renaming a few variables.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
+++ b/src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
@@ -1,5 +1,10 @@
 package uk.gov.pay.directdebit.notifications.services;
 
+import com.google.common.collect.ImmutableMap;
+import java.math.BigDecimal;
+import java.time.LocalDate;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
```diff
@@ -36,45 +34,40 @@ public UserNotificationService(AdminUsersClient adminUsersClient, DirectDebitCon
...
+                        .build());
     }
 
     public void sendPaymentFailedEmailFor(Transaction transaction, Payer payer) {
-        adminUsersClient.sendEmail(EmailTemplate.PAYMENT_FAILED, transaction, payer.getEmail(), buildPersonalisation());
-    }
-
-    private HashMap<String, String> buildMandateProblemPersonalisation(Mandate mandate) {
-        HashMap<String, String> map = new HashMap<>();
-        // fixme use the right reference once we play PP-3547
```
alexbishop1: You fixed it!

PR #156, opened 2018-06-01
PP-3776 Expose 50505 port to be used with Java Debug Wire Protocol (JDWP)

DanailMinchev:


mrlumbu:
Can you include in commit message and PR why you need this (ie. explain you are trying to debug a specific problem). That way our future selves will know whether we can remove it or not

PR #159, opened 2018-06-06
PP-3676 Refactor data model to allow on-demand agreements 

simad:
## WHAT
- The nexus of our data model was payment request. This is not valid anymore when introducing on-demand agreements, as it is possible to create a mandate which is not linked to any payment.
- Refactoring the data model so that mandate is now at its center. It is also now possible to have multiple transactions per mandate.
- Removed payment request in favour of transaction

### DIFF
--- a/src/main/resources/migrations/00019_alter_transactions_requests_mandate_id.sql
+++ b/src/main/resources/migrations/00019_alter_transactions_requests_mandate_id.sql
@@ -17,7 +17,7 @@ ALTER TABLE transactions ADD COLUMN mandate_id BIGINT;
 --rollback ALTER TABLE transactions DROP COLUMN mandate_id;
 
 --changeset uk.gov.pay:alter_table-transactions-external_id
-ALTER TABLE transactions ADD COLUMN external_id VARCHAR(30);
+ALTER TABLE transactions ADD COLUMN external_id VARCHAR(26);
 --rollback ALTER TABLE transactions DROP COLUMN external_id;
 
 --changeset uk.gov.pay:alter_table-transactions-reference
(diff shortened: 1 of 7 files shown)


src/main/resources/migrations/00019_alter_transactions_requests_mandate_id.sql
```diff
@@ -0,0 +1,35 @@
...
+--changeset uk.gov.pay:alter_table-payers-payment-requests-fk
+ALTER TABLE transactions DROP CONSTRAINT transactions_payment_requests_fk
+--rollback add constraint transactions_payment_requests_fk FOREIGN KEY (payment_request_id) REFERENCES payment_requests (id);
+
+--changeset uk.gov.pay:alter_table-transactions-mandate-id
+ALTER TABLE transactions ADD COLUMN mandate_id BIGINT;
+--rollback ALTER TABLE transactions DROP COLUMN mandate_id;
+
+--changeset uk.gov.pay:alter_table-transactions-external_id
+ALTER TABLE transactions ADD COLUMN external_id VARCHAR(30);
```
DanailMinchev: Can `external_id` be `VARCHAR(26)` if you are using https://github.com/alphagov/pay-direct-debit-connector/blob/b39cef769e099ece3ff974c585375c4aa85688b9/src/main/java/uk/gov/pay/directdebit/common/util/RandomIdGenerator.java#L21 ?

simad: Yes, I will make sure mandate and transaction external id are both VARCHAR(26)

PR #164, opened 2018-06-15
Publicapi contract test

oswaldquek:
Provider test for interaction between publicapi and direct-debit-connector.

### DIFF
--- a/src/test/java/uk/gov/pay/directdebit/junit/DropwizardAppWithPostgresRule.java
+++ b/src/test/java/uk/gov/pay/directdebit/junit/DropwizardAppWithPostgresRule.java
@@ -20,7 +20,6 @@ import static uk.gov.pay.directdebit.junit.PostgresTestDocker.getDbUri;
 import static uk.gov.pay.directdebit.junit.PostgresTestDocker.getOrCreate;
 
 public class DropwizardAppWithPostgresRule implements TestRule {
-    private static final Logger logger = LoggerFactory.getLogger(DropwizardAppWithPostgresRule.class);
 
     private final DropwizardAppRule<DirectDebitConfig> app;
     private final String configFilePath;
--- a/src/test/java/uk/gov/pay/directdebit/junit/DropwizardAppWithPostgresRule.java
+++ b/src/test/java/uk/gov/pay/directdebit/junit/DropwizardAppWithPostgresRule.java
@@ -1,0 +1,62 @@
+package uk.gov.pay.directdebit.junit;
+
+import io.dropwizard.testing.ConfigOverride;
+import io.dropwizard.testing.junit.DropwizardAppRule;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.directdebit.DirectDebitConnectorApp;
+import uk.gov.pay.directdebit.app.config.DirectDebitConfig;
+
+import static io.dropwizard.testing.ConfigOverride.config;
+import static io.dropwizard.testing.ResourceHelpers.resourceFilePath;
+import static uk.gov.pay.directdebit.junit.PostgresTemplate.createTemplate;
+import static uk.gov.pay.directdebit.junit.PostgresTestContainer.DB_PASSWORD;
+import static uk.gov.pay.directdebit.junit.PostgresTestContainer.DB_USERNAME;
+import static uk.gov.pay.directdebit.junit.PostgresTestDocker.getDbUri;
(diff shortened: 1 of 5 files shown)


oswaldquek:
DO NOT MERGE

src/test/java/uk/gov/pay/directdebit/pact/PublicApiContractTest.java
```diff
@@ -0,0 +1,43 @@
...
+    @BeforeClass
+    public static void setUpService() {
+        target = new HttpTarget(app.getLocalPort());
+    }
+    
+    @State("a gateway account with external id exists")
+    public void aGatewayAccountWithExternalIdExists(Map<String, String> params) {
+        testGatewayAccount.withExternalId(params.get("gateway_account_id")).insert(app.getTestContext().getJdbi());
+    }
+}
```
mrlumbu: It would be good to find some way of being able to have separate pact test files for different bits of functionality, otherwise this file will become huge and unmanageable.

src/test/java/uk/gov/pay/directdebit/pact/PublicApiContractTest.java
```diff
@@ -0,0 +1,43 @@
...
+    public static Target target;
+
+    private GatewayAccountFixture testGatewayAccount = GatewayAccountFixture.aGatewayAccountFixture();
+
+    @BeforeClass
+    public static void setUpService() {
+        target = new HttpTarget(app.getLocalPort());
+    }
+    
+    @State("a gateway account with external id exists")
```
mrlumbu: I think we should specify the id in the provider state string (I realise this would need changes in both publicapi and here). The more specific and informative we can make these state strings, the less confused we'll be, I hope

src/test/java/uk/gov/pay/directdebit/junit/DropwizardAppWithPostgresRule.java
```diff
@@ -0,0 +1,61 @@
+package uk.gov.pay.directdebit.junit;
+
+import io.dropwizard.testing.ConfigOverride;
+import io.dropwizard.testing.junit.DropwizardAppRule;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
mrlumbu: these are unused

src/test/java/uk/gov/pay/directdebit/pact/PublicApiContractTest.java
oswaldquek: My view is that using the `Map<String, String> params` would be better as this would allow code reuse. If you have a look at https://github.com/alphagov/pay-publicapi/blob/BAU-add-pact-testing/src/test/resources/pacts/publicapi-publicauth-direct-debit.json#L11 for example it defines a state "A bearer token exists" that provides a "type" and "id". I can easily see another pact being created that needs the same but with a different "type" (CARD for example). This means we'd only have to define one `@State` method in the provider test.

src/test/java/uk/gov/pay/directdebit/pact/PublicApiContractTest.java
oswaldquek: this was one of the problems at my last job! The provider test class was getting huge. I don't know if there's a way to separate it out just now. I'll ask in the pact slack but I don't think it's an issue that should prevent this PR from progressing at the moment as the provider test class still looks quite clean.

src/test/java/uk/gov/pay/directdebit/pact/PublicApiContractTest.java
mrlumbu: oh that's cool, I didn't know you could do that

oswaldquek: yeah one of the ways pact-jvm is better than the JS one :)

src/test/java/uk/gov/pay/directdebit/junit/DropwizardAppWithPostgresRule.java
oswaldquek: fixed

src/test/java/uk/gov/pay/directdebit/pact/PublicApiContractTest.java
mrlumbu: ok cool. Would be good to figure this out soonish

src/test/java/uk/gov/pay/directdebit/pact/PublicApiContractTest.java
mrlumbu: 👍

PR #168, opened 2018-06-19
PP-3676 Cleanup migrations after updating the data model (part 1)

simad:
## WHAT
- Removing payment_request_id from tables, as we now assume there's
  a mandate_id to link tables together
- Drop table payment_requests
- Drop constraint NOT_NULL for type of transaction, this will be dropped in another PR
- Adding a new table events, will drop payment_request_events in another PR

## HOW 
_Steps to test or reproduce:_


heathd:
looks fine

PR #169, opened 2018-06-19
PP-3676 Add internal_event_id in gocardless events pointing to the new events table

simad:
## WHAT
- Add `internal_event_id` to gocardless event and make it reference the new `events` table. `payment_request_event_id` will be dropped in the next PR

### DIFF
--- a/src/main/resources/migrations/00029_alter_table_gocardless_events.sql
+++ b/src/main/resources/migrations/00029_alter_table_gocardless_events.sql
@@ -1,0 +1,13 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:alter_table-gocardless-events-internal-id
+ALTER TABLE gocardless_events ADD COLUMN internal_event_id BIGINT;
+--rollback DROP COLUMN internal_event_id;
+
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/00029_alter_table_gocardless_events.sql
```diff
@@ -0,0 +1,13 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:alter_table-gocardless-events-internal-id
+ALTER TABLE gocardless_events ADD COLUMN internal_event_id BIGINT;
```
heathd: why not call it `event_id`?

heathd:
[minor nitpick] would be good to include the same explanation in the commit that you put in the pull request message, since that's more easily accessible to future developers looking at git history.

heathd:
Happy to merge this as-is, although one small question

src/main/resources/migrations/00029_alter_table_gocardless_events.sql
simad: because there is already an event_id which is currently used as a gocardless_event_id :(

PR #171, opened 2018-06-19
PP-3676 Drop type from transactions and drop payment_request_events

simad:
## WHAT
- transaction type is not used (it's always charge) 
- we are using a table called `events` now, so `payment_request_events` can be dropped

### DIFF
--- a/src/main/resources/migrations/00030_alter_table_transactions_drop_type.sql
+++ b/src/main/resources/migrations/00030_alter_table_transactions_drop_type.sql
@@ -1,0 +1,5 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:alter_table-transactions-drop-type
+ALTER TABLE transactions DROP COLUMN type;
+--rollback ADD COLUMN type TEXT;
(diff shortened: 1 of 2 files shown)


src/main/resources/migrations/00030_alter_table_transactions_drop_type.sql
```diff
@@ -0,0 +1,5 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:alter_table-transactions-drop-type
+ALTER TABLE transactions DROP COLUMN type;
```
alexbishop1: 💣

src/main/resources/migrations/00031_drop_table_payment_request_events.sql
```diff
@@ -0,0 +1,5 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:drop-table-payment-request-events
+DROP TABLE payment_request_events;
```
alexbishop1: 💥

PR #173, opened 2018-06-20
PP-3676 Completely eradicate payment request from the codebase

simad:
## WHAT
- Renamed everything which has `PaymentRequest` as prefix to Transaction, wherever it does make sense
- `Event` is a bit too generic and we decided to rename it to `DirectDebitEvent`


heathd:
good move, but I'm not convinced about renaming `Event` to `DirectDebitEvent`. 

within the context of direct debit connector, what other kind of events exist?

simad:
That came up as one of the original pr review comments, it's not really adding anything "semantically" but it's just to disambiguate Event from the 1929393 java classes with the same name. If you can think of a better name, suggestions are more than welcome :)

PR #175, opened 2018-06-21
PP-3676 Update mandates table schema to handle references

DanailMinchev:
## WHAT

- Added `mandate_reference` column which will replace `reference` column in the next releases.
  This is necessary to distinguish between mandate reference and the future service reference.
- Added `service_reference` column which will hold optional service reference value.

with @georgeracu


oswaldquek:
LGTM 🐼

PR #177, opened 2018-06-21
PP-3676 Renamed Mandate#reference to Mandate#mandateReference

DanailMinchev:
## WHAT

- Renamed `Mandate`'s `reference` field to `mandateReference` on resource and database level
- Added `Mandate#serviceReference`
- Minor code styling corrections and refactoring

with @georgeracu


heathd:
we didn't change the field exposed in the token response. It's still called `reference`. That might get confusing. Is there any reason not to change it to `mandate_reference`?

heathd:
Looks fine.

Couple of things we might want to do in a subsequent PR:

1. add a description of the API in a `docs/` folder, either as markdown file or open api specification format
2. change the return value of the token api to rename `reference` to `mandate_reference`

DanailMinchev:
Thank you @heathd , we will address that in the next PRs

PR #182, opened 2018-06-22
PP-3679 Refactor search transactions to include mandate_id

SandorArpa:
Payment view should be able to be searched by a mandate id. Search response should also mirror to a logical extent that of connector search, eg paginations, total and count.

- add mandate_id and and query string generation to `PaymentViewSearchParams`
- refactor `PaymentViewSearchParams` to use `withXXXParam` instead of an unreadable constructor
- add `withUriInfo` to `PaymentViewService` to enable the service to generate links
- introduce `Link` class
- add feature to searching to include the total number of records for a given search
- refactor searching to include the number of records on the given page and the page number
- add `ViewPaginationBuilder` (to mirror connector's search pagination)
- add relevant Pact test
- add relevant tests

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/payments/dao/PaymentViewDao.java
+++ b/src/main/java/uk/gov/pay/directdebit/payments/dao/PaymentViewDao.java
@@ -2,6 +2,8 @@ package uk.gov.pay.directdebit.payments.dao;
 
 import java.util.List;
 import javax.inject.Inject;
+import javax.ws.rs.QueryParam;
+
 import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.core.statement.Query;
 import uk.gov.pay.directdebit.payments.dao.mapper.PaymentViewMapper;
@@ -11,7 +13,7 @@ import uk.gov.pay.directdebit.payments.params.PaymentViewSearchParams;
 public class PaymentViewDao {
     private final Jdbi jdbi;
     private final String QUERY_STRING = "SELECT ga.external_id as gateway_external_id, " +
-            "t.external_id AS payment_request_id, " +
+            "t.external_id AS transaction_external_id, " +
             "t.amount AS amount, " +
             "t.reference AS reference, " +
             "t.description AS description, " +
@@ -22,10 +24,18 @@ public class PaymentViewDao {
             "FROM transactions t " +
             "INNER JOIN mandates m ON t.mandate_id = m.id " +
             "INNER JOIN gateway_accounts ga ON ga.id = m.gateway_account_id " +
-            "LEFT JOIN payers pa ON t.mandate_id = pa.mandate_id " +
+            "INNER JOIN payers pa ON m.id = pa.mandate_id " +
             "WHERE ga.external_id = :gatewayAccountExternalId " +
             ":searchExtraFields " +
             "ORDER BY t.id DESC OFFSET :offset LIMIT :limit";
+
+    private final String COUNT_QUERY_STRING = "SELECT count(*) " +
+            "FROM transactions t " +
+            "INNER JOIN mandates m ON t.mandate_id = m.id " +
+            "INNER JOIN gateway_accounts ga ON ga.id = m.gateway_account_id " +
+            "INNER JOIN payers pa ON m.id = pa.mandate_id " +
+            "WHERE ga.external_id = :gatewayAccountExternalId " +
+            ":searchExtraFields ";
 
     @Inject
     public PaymentViewDao(Jdbi jdbi) {
@@ -40,6 +50,17 @@ public class PaymentViewDao {
             return query
                     .map(new PaymentViewMapper())
                     .list();
+        });
+    }
+    
+    public Long getPaymentViewCount(PaymentViewSearchParams searchParams) {
+        String searchExtraFields = searchParams.generateQuery();
+        return jdbi.withHandle(handle -> {
+            Query query = handle.createQuery(COUNT_QUERY_STRING.replace(":searchExtraFields", searchExtraFields));
+            searchParams.getQueryMap().forEach(query::bind);
+            return query
+                    .mapTo(Long.class)
+                    .findOnly();
         });
     }
 }
--- a/src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
+++ b/src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
@@ -3,6 +3,7 @@ package uk.gov.pay.directdebit.payments.params;
 import java.util.HashMap;
 import java.util.Map;
 
+import static java.lang.String.format;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
 public class PaymentViewSearchParams {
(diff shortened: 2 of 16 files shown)


src/main/java/uk/gov/pay/directdebit/payments/dao/PaymentViewDao.java
```diff
@@ -27,6 +27,14 @@
             ":searchExtraFields " +
             "ORDER BY t.id DESC OFFSET :offset LIMIT :limit";
 
+    private final String COUNT_QUERY_STRING = "SELECT count(*) " +
+            "FROM transactions t " +
+            "INNER JOIN mandates m ON t.mandate_id = m.id " +
+            "INNER JOIN gateway_accounts ga ON ga.id = m.gateway_account_id " +
+            "LEFT JOIN payers pa ON t.mandate_id = pa.mandate_id " +
```
mrlumbu: out of interest do you know why this is a left join? Can there be a payment without a payer?

SandorArpa: In theory there shouldn't be a payment without payer. I just stripped down the search sql and to be honest I didn't pay attention to the left join. I shall remove it.

src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
```diff
@@ -72,14 +109,25 @@ public PaginationParams getPaginationParams() {
...
+        return this;
+    }
 
     public String generateQuery() {
         StringBuilder sb = new StringBuilder("");
-        if (searchDateParams.getFromDate() != null) {
-            sb.append(" AND t.created_date > :" + FROM_DATE_FIELD);
+        if (isNotBlank(mandateId)) {
+            sb.append(" AND m.external_id = :" + MANDATE_EXTERNAL_ID_FIELD);
+            return sb.toString();
```
mrlumbu: can you explain this? I don't think I understand why we ignore all the other params if mandate id set?

src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
```diff
@@ -99,11 +147,17 @@ public String generateQuery() {
             queryMap.put(GATEWAY_ACCOUNT_EXTERNAL_FIELD, gatewayExternalId);
             queryMap.put(PAGE_NUMBER_FIELD, getPaginationParams().getPageNumber());
             queryMap.put(PAGE_SIZE_FIELD, getPaginationParams().getDisplaySize());
-            if (searchDateParams.getFromDate() != null) {
-                queryMap.put(FROM_DATE_FIELD, searchDateParams.getFromDate());
+            if (isNotBlank(mandateId)) {
+                queryMap.put(MANDATE_EXTERNAL_ID_FIELD, mandateId);
+                return queryMap;
```
mrlumbu: likewise here

src/main/java/uk/gov/pay/directdebit/payments/resources/PaymentViewResource.java
```diff
@@ -41,17 +44,21 @@ public Response getPaymentView(
...
+            @Context UriInfo uriInfo){
+        PaymentViewSearchParams searchParams = new PaymentViewSearchParams(accountExternalId)
+                .withPage(pageNumber)
+                .withDisplaySize(displaySize)
+                .withFromDateString(fromDate)
+                .withToDateString(toDate)
+                .withEmail(email)
+                .withReference(reference)
+                .withAmount(amount)
+                .withMandateId(mandateId);
```
mrlumbu: like the new builders you've made 👍

src/main/java/uk/gov/pay/directdebit/payments/services/PaymentViewService.java
```diff
@@ -31,26 +40,46 @@ public PaymentViewResponse getPaymentViewResponse(PaymentViewSearchParams search
         return gatewayAccountDao.findByExternalId(searchParams.getGatewayExternalId())
                 .map(gatewayAccount -> {
                     PaymentViewSearchParams validatedSearchParams = paymentViewValidator.validateParams(searchParams);
-                    List<PaymentViewListResponse> viewListResponse = 
-                            getPaymentViewListResponse(paymentViewValidator.validateParams(searchParams));
+                    Long total = getTotal(validatedSearchParams);
+                    if (total == 0) {
+                        throw new RecordsNotFoundException(format("Found no records for %s", validatedSearchParams.buildQueryParamString()));
```
mrlumbu: I don't understand why we're throwing an exception if we don't find anything? THis is a normal state of affairs, and we should just return and empty array no? (I realise you didn't initiate this pattern)

src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
SandorArpa: In the requirement we explicitly say mandate id (it was not added as an extra search param to PP-3646). It was originally thought to be a different search to that of transactions search. I combined the two as both should return the same search response. At the kick off time it was decided that we ignore the other search results for now. It was also discussed that ignoring other params feature is going to be removed at a later stage.

src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
SandorArpa: See above

src/main/java/uk/gov/pay/directdebit/payments/services/PaymentViewService.java
SandorArpa: Good point, I shall remove it.

src/main/java/uk/gov/pay/directdebit/payments/services/PaymentViewService.java
```diff
@@ -31,26 +38,42 @@ public PaymentViewResponse getPaymentViewResponse(PaymentViewSearchParams search
...
-                            getPaymentViewListResponse(paymentViewValidator.validateParams(searchParams));
+                    Long total = getTotal(validatedSearchParams);
+                    List<PaymentViewListResponse> viewListResponse =
+                            getPaymentViewListResponse(validatedSearchParams, gatewayAccount.getExternalId());
                     if (viewListResponse.size() == 0) {
                         throw new RecordsNotFoundException(format("Found no records with page size %s and display_size %s",
                                 validatedSearchParams.getPage(),
                                 validatedSearchParams.getDisplaySize()));
                     }
+                    ViewPaginationBuilder paginationBuilder = new ViewPaginationBuilder(validatedSearchParams, viewListResponse, uriInfo);
```
mrlumbu: I think you should remove the other exception as well. Also you could do 
```
if (total != 0) {
    getPaymentViewListResponse...
```
i.e. don't make do the search if you know there are no records

src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
mrlumbu: I'm not convinced by this, and I think it could become rather confusing. I think there is a mixing up of different concepts here: 
- the idea of payments as a subresource of mandate (i.e. `/mandate/:mandateId/payments)
- the idea of a search for payments matching certain query params (i.e. `/payments?mandate=:mandateId`)
You seem to be implementing the first concept using the tools of the second concept. I don't think I am qualified to decide if this is the way to go in this project (I personally don't think it is). I realise in publicapi there has been a deliberate decision to do it this way, but I am not sure that means we need to persist it all the way through

SandorArpa: I don't know what to say to that.

src/test/java/uk/gov/pay/directdebit/payments/dao/PaymentViewDaoIT.java
```diff
@@ -214,9 +223,33 @@ public void shouldReturnOnePaymentViewWhenPaymentCreated() {
...
+                    .withDescription("description " + i)
+                    .withAmount(1000L + i)
+                    .insert(testContext.getJdbi());
+        }
+        PaymentViewSearchParams searchParams = new PaymentViewSearchParams(gatewayAccountFixture.getExternalId())
+                .withMandateId(mandateFixture.getExternalId());
+        Long count = paymentViewDao.getPaymentViewCount(searchParams);
+        assertThat(count, is(3L));
+    }
 }
```
mrlumbu: think you need one more test in here just to prove that searching by mandate id works - ie have two payments with different mandate ids and search for payment by mandate id and check you get the right one back

src/test/java/uk/gov/pay/directdebit/payments/resources/PaymentViewResourceITest.java
```diff
@@ -341,15 +308,15 @@ public void shouldReturn6Records_whenAmountSet() {
...
         givenSetup()
                 .get(requestPath)
                 .then()
                 .statusCode(Response.Status.OK.getStatusCode())
                 .contentType(JSON)
-                .body("payment_views", hasSize(6));
+                .body("results", hasSize(6));
     }
 
     private RequestSpecification givenSetup() {
```
mrlumbu: think you need a test for mandate id in here too?

src/test/java/uk/gov/pay/directdebit/payments/dao/PaymentViewDaoIT.java
```diff
@@ -214,9 +226,76 @@ public void shouldReturnOnePaymentViewWhenPaymentCreated() {
...
+        PaymentViewSearchParams searchParams = new PaymentViewSearchParams(gatewayAccountFixture.getExternalId())
+                .withMandateId(mandateExternalId);
+        Long count = paymentViewDao.getPaymentViewCount(searchParams);
+        assertThat(count, is(3L));
+        List<PaymentView> paymentViews = paymentViewDao.searchPaymentView(searchParams);
+        assertThat(paymentViews.size(), is(3));
+        assertThat(paymentViews.get(0).getEmail().equals(payerFixture1.getEmail()), is(true));
+        assertThat(paymentViews.get(1).getEmail().equals(payerFixture1.getEmail()), is(true));
+        assertThat(paymentViews.get(2).getEmail().equals(payerFixture1.getEmail()), is(true));
+    }
```
mrlumbu: 👍

src/test/java/uk/gov/pay/directdebit/payments/resources/PaymentViewResourceITest.java
SandorArpa: Mandate id is not included in the response json object

PR #186, opened 2018-06-26
PP-3679 Change endpoint queryparam from mandate_id to agreement

SandorArpa:
this change is needed to hide mandate_id from public api

- change mandate_id to agreement on `PaymentViewResource` and `PaymentViewSearchParams.buildQueryParamString()`
- update relevant test

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
+++ b/src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
@@ -18,6 +18,7 @@ public class PaymentViewSearchParams {
     private static final String AMOUNT_FIELD = "amount";
     private static final String MANDATE_EXTERNAL_ID_FIELD = "mandate_external_id";
     private static final String MANDATE_ID_KEY = "mandate_id";
+    private static final String MANDATE_ID_EXTERNAL_KEY = "agreement";
     private static final String FROM_DATE_KEY = "from_date";
     private static final String TO_DATE_KEY = "to_date";
     private final String gatewayExternalId;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
```diff
@@ -173,7 +174,7 @@ public String generateQuery() {
     public String buildQueryParamString() {
         StringBuilder sb = new StringBuilder("");
         if (isNotBlank(mandateId)) {
-            sb.append("&" + MANDATE_ID_KEY + "=" + mandateId);
+            sb.append("&" + MANDATE_ID_EXTERNAL_KEY + "=" + mandateId);
```
alexbishop1: You’ve already got a `StringBuilder` here. Using `+` concatenation with a variable that isn’t a constant will create another one.

src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
```diff
@@ -171,29 +172,29 @@ public String generateQuery() {
...
-            sb.append("&" + REFERENCE_FIELD + "=" + reference);
+            query += "&" + REFERENCE_FIELD + "=" + reference;
         }
         if (amount != null) {
-            sb.append("&" + AMOUNT_FIELD + "=" + amount);
+            query += "&" + AMOUNT_FIELD + "=" + amount;
         }
-        sb.append(addPaginationParams());
-        return sb.toString();
+        query += addPaginationParams();
```
alexbishop1: If I’m remembering my Java internals correctly, this will cause lots of `StringBuilder`s to be created: the compiler isn’t smart enough to use the same `StringBuilder` across multiple statements.

Where we have lots of conditionals mixed in, it’s best to create a `StringBuilder` and then always use `append` to add to it (e.g. `sb.append("&").append(MANDATE_ID_KEY).append("=").append(mandateId)` etc.).

Not that any of this is likely to make a huge amount of difference to performance!

PR #188, opened 2018-06-27
PP-3676 Fix mandate reference generated when we use sandbox account

DanailMinchev:
## WHAT

- We added default value placeholders for both Sandbox and GoCardless

with @georgeracu

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/services/MandateService.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/services/MandateService.java
@@ -1,17 +1,7 @@
 package uk.gov.pay.directdebit.mandate.services;
 
 import com.google.common.collect.ImmutableMap;
-
-import java.time.ZoneOffset;
-import java.time.ZonedDateTime;
-import java.util.ArrayList;
-import java.util.List;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/directdebit/mandate/services/MandateService.java
```diff
@@ -90,7 +90,10 @@ public Mandate createMandate(Map<String, String> mandateRequestMap, String accou
                             gatewayAccount,
                             MandateType.valueOf(mandateRequestMap.get("agreement_type")),
                             RandomIdGenerator.newId(),
-                            null,
+                            // TODO:
+                            // when we introduce GoCardless gateway accounts to work with create mandate,
+                            // then modify appropriate mandate reference values
+                            PaymentProvider.SANDBOX.equals(gatewayAccount.getPaymentProvider()) ? "sandbox-default" : "gocardless-default",
```
alexbishop1: For sandbox, it would be more realistic to generate the mandate reference manually

Fine with using something static for GoCardless because we will be using mandate references generated by GoCardless by the time anyone starts using this

PR #196, opened 2018-06-29
PP-3679 refactor to accept both agreement and agreement_id as queryparams part 1

SandorArpa:
- refactor Pact test to try to use mandate_id by either agreement or agreement_id
- refactor PaymentViewResource to accept both agreement and agreement_id


heathd:
looks fine

I'm not keen on the `_BWC` name as it's a bit opaque. I'm not really sure these fields need to be constants at all, since they are only used in one place.

I'd be inclined to just inline them all ie.

```
    public Response getPaymentView(
            @PathParam("accountId") String accountExternalId,
...
            @QueryParam("agreement") String mandateId,
            @QueryParam("agreement_id") String mandateBWCId,
            @Context UriInfo uriInfo){
```

since this is a short-lived backward compatibility I think it's fine for now.

PR #198, opened 2018-06-29
PP-3671 Renamed type to mandate type in Token response

DanailMinchev:
with @georgeracu

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/tokens/api/TokenResponse.java
+++ b/src/main/java/uk/gov/pay/directdebit/tokens/api/TokenResponse.java
@@ -12,9 +12,6 @@ public class TokenResponse {
     @JsonProperty("external_id")
     private String mandateExternalId;
 
-    @JsonProperty("type")
-    private String type;
-
     @JsonProperty("state")
     private String state;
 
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/directdebit/tokens/api/TokenResponse.java
```diff
@@ -30,22 +27,24 @@
     @JsonProperty("mandate_reference")
     private String mandateReference;
 
+    @JsonProperty("mandate_type")
```
alexbishop1: What are the implications of this change?

DanailMinchev: Currently frontend is not using this property (it is just parsed) here - https://github.com/alphagov/pay-direct-debit-frontend/blob/b3f29b12228874c7d363c964ed995e16d8ab64e3/common/clients/connector-client.js#L43
and
https://github.com/alphagov/pay-direct-debit-frontend/blob/b3f29b12228874c7d363c964ed995e16d8ab64e3/common/classes/Mandate.class.js#L14
so we renamed it to be consistent with:
- https://github.com/alphagov/pay-direct-debit-connector/blob/master/src/main/java/uk/gov/pay/directdebit/mandate/api/CreateMandateResponse.java#L24
- https://github.com/alphagov/pay-direct-debit-connector/blob/master/src/main/java/uk/gov/pay/directdebit/mandate/api/DirectDebitInfoFrontendResponse.java#L123
- https://github.com/alphagov/pay-direct-debit-connector/blob/master/src/main/java/uk/gov/pay/directdebit/mandate/api/GetMandateResponse.java#L20

and in our next frontend PR we are going to use the new property

alexbishop1: OK, if it’s not been used yet, that’s fine

PR #199, opened 2018-06-29
Pp 3680 events pagination

danworth:
Improved the pagination for event search to include links to first, last, next, self and previous pages. This also introduces "count" as the number of results in the returned page and "total"  as the total number of results available for the parameters provided (e.g. across all pages).

This follows the pattern implemented for Payments search. 

@oswaldquek

### DIFF
--- a/src/test/java/uk/gov/pay/directdebit/events/api/DirectDebitEventsPaginationTest.java
+++ b/src/test/java/uk/gov/pay/directdebit/events/api/DirectDebitEventsPaginationTest.java
@@ -13,6 +13,7 @@ import javax.ws.rs.core.UriBuilder;
 import javax.ws.rs.core.UriInfo;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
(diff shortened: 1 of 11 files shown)


src/test/java/uk/gov/pay/directdebit/events/api/DirectDebitEventsPaginationTest.java
```diff
@@ -0,0 +1,93 @@
...
+@RunWith(MockitoJUnitRunner.class)
+public class DirectDebitEventsPaginationTest {
+
+    
+    @Mock
+    private UriInfo mockUriInfo;
+
+    @Before
+    public void setup() {
+        when(mockUriInfo.getBaseUriBuilder()).thenAnswer((Answer<UriBuilder>) invocation -> UriBuilder.fromUri("http://test.com"));
```
alexbishop1: The domain name `test.com` is a real, registered domain name. Please use something like `whateveryoulike.test` — `.test` domains are guaranteed never to be registered (see RFC 2606)

oswaldquek:
@alexbishop1 could you add some details as to how it works in a different fashion? I can see ViewPaginationBuilder and DirectDebitEventsPagination are probably good candidates for code reuse but other than that the approaches largely the same to me.

alexbishop1:
@oswaldquek It could just be my ignorance of how the existing search stuff works showing, sorry. But I couldn’t see equivalents of, say, `DirectDebitEventSearchParams`. I thought the analogues would be clearer. But this could just be a misunderstanding on my part. The code you have written looks good.

oswaldquek:
@alexbishop1 there's a `PaymentViewSearchParams` :) We largely copied the structure from https://github.com/alphagov/pay-direct-debit-connector/blob/e68d66c915421c774fe12f0d8c301dcf930beeb2/src/main/java/uk/gov/pay/directdebit/payments/resources/PaymentViewResource.java, where it calls a `PaymentViewService` which calls some DAO.

PR #203, opened 2018-07-03
PP-3676 Added missing template keys for ON_DEMAND mandate created email

DanailMinchev:
with @georgeracu

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
+++ b/src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
@@ -14,7 +14,8 @@ import java.time.format.DateTimeFormatter;
 
 public class UserNotificationService {
 
-    private static final String PLACEHOLDER_STATEMENT_NAME = "THE-CAKE-IS-A-LIE";
+    public static final String PLACEHOLDER_STATEMENT_NAME = "THE-CAKE-IS-A-LIE";
+    public static final String BANK_ACCOUNT_MASK_PREFIX = "******";
 
     private static final String DD_GUARANTEE_KEY = "dd guarantee link";
     private static final String MANDATE_REFERENCE_KEY = "mandate reference";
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
```diff
@@ -14,7 +14,8 @@
 
 public class UserNotificationService {
 
-    private static final String PLACEHOLDER_STATEMENT_NAME = "THE-CAKE-IS-A-LIE";
+    public static final String PLACEHOLDER_STATEMENT_NAME = "THE-CAKE-IS-A-LIE";
```
mrlumbu: is this a good idea?

DanailMinchev: @mrlumbu we didn't want to copy the value, that is why we made it `public`

mrlumbu: OK clarified where`THE-CAKE-IS-A-LIE` fits into the whole picture. Ultimately this will be supplied by adminusers. This will be cleaned up as part of service onboarding stories. Hopefully 😬

src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
```diff
@@ -44,10 +45,12 @@ public void sendMandateFailedEmailFor(Mandate mandate) {
 
     public void sendOnDemandMandateCreatedEmailFor(Mandate mandate) {
         adminUsersClient.sendEmail(EmailTemplate.ON_DEMAND_MANDATE_CREATED, mandate,
-                ImmutableMap.of(
-                        MANDATE_REFERENCE_KEY, mandate.getMandateReference(),
-                        DD_GUARANTEE_KEY, directDebitConfig.getLinks().getDirectDebitGuaranteeUrl()
-                )
+                ImmutableMap.<String, String>builder()
+                        .put(MANDATE_REFERENCE_KEY, mandate.getMandateReference())
+                        .put(BANK_ACCOUNT_LAST_DIGITS_KEY, BANK_ACCOUNT_MASK_PREFIX + mandate.getPayer().getAccountNumberLastTwoDigits())
```
mrlumbu: why do we have to send a load of asterisks. can't that just be in the template? Not a big deal

DanailMinchev: There was a bug in our email provider before:
If we define more than two asterisks in the template it was truncated to two, that is why we are passing the asterisks from our code.
Katie will verify if the bug still exists and we can address the changes in the next PR.

PR #207, opened 2018-07-04
PP-3942 Introduce AccountNumber

alexbishop1:
The `AccountNumber` object represents a UK bank account number with no dashes or spaces e.g. `"12345678"`. It performs a check when it’s constructed to ensure that it consists of six to eight Arabic numerals inclusive. It’s safe to do this because `AccountNumber` objects are only ever created to represent data in requests and our resources ensure the account numbers match the correct format before the `AccountNumber` objects are created.

Note that there’s no guarantee that an `AccountNumber` object actually represents a real bank account.

The `accountNumber` field in the `Payer` class remains a `String` rather than becoming an `AccountNumber` object because we hash the account number before passing it to the `Payer` constructor (except in some tests), so it isn’t actually an account number. This also means that `AccountNumber` objects are never persisted or depersisted: they are entirely ephemeral.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/payers/model/AccountNumber.java
+++ b/src/main/java/uk/gov/pay/directdebit/payers/model/AccountNumber.java
@@ -1,0 +1,53 @@
+package uk.gov.pay.directdebit.payers.model;
+
+import java.util.Objects;
+import java.util.regex.Pattern;
+
+/**
+ * A UK bank account number with no dashes or spaces e.g "12345678"
+ * Guaranteed to be well-formed (six to eight Arabic numerals
+ * inclusive) but not necessarily valid (in the sense of representing
+ * a real bank account)
+ */
+public class AccountNumber {
+
+    private static final Pattern SIX_TO_EIGHT_ARABIC_NUMERALS = Pattern.compile("[0-9]{6,8}");
+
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/pay/directdebit/payers/model/AccountNumber.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * A UK bank account number with no dashes or spaces e.g "12345678"
+ * Guaranteed to be well-formed (six to eight Arabic numerals
+ * inclusive) but not necessarily valid (in the sense of representing
+ * a real bank account)
+ */
+public class AccountNumber {
+
+    private static final Pattern SIX_TO_EIGHT_ARABIC_NUMERALS = Pattern.compile("[0-9]{6,8}");
```
simad: I am happy to immediately have the information that they are not roman numerals. Good job

alexbishop1: You have to be careful! In some languages, `\d` in a regex matches more than just the Arabic numerals 0–9: https://blog.mariusschulz.com/2015/10/21/matching-digits-in-net-regex

PR #210, opened 2018-07-06
Pp 3942 introduce `MandateExternalId`

georgeracu:
Replace the `String` with `MandateExternalId`

- regsitered an argumentfactory to convert MandateExternalId to String for JDBI
- Added a new type MandateExternalId that wraps a String value, used to
represent an external id for a Mandate;
- Added `valueOf` for automatic conversion into object at resource level
- Added getter for ObjectWrapper to be able to convert into JSON in tests
- Added custom ArgumentFactory for jdbi to be able to map this new type
to the database CHAR
- Registered the custom ArgumentFactory for DAOs and for test context

with @mrlumbu, @simad, @georgeracu, @kbottla, @heathd, @oswaldquek,
@DanailMinchev, @danworth, @schmonz (we were mob-programming!)


heathd:
I'm working on rebasing this

src/main/java/uk/gov/pay/directdebit/mandate/model/subtype/MandateExternalId.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private MandateExternalId(String mandateExternalId) {
+        this.mandateExternalId = Objects.requireNonNull(mandateExternalId);
+    }
+
+    public static MandateExternalId of(String mandateExternalId) {
+        return new MandateExternalId(mandateExternalId);
+    }
+
+    public static MandateExternalId valueOf(String mandateExternalId) {
```
danworth: Do we need `valueOf` in addition to the `of` method, I don't see it used anywhere?

heathd:
regarding `valueOf`, it's used by JAX-RS's serialisation:

> JAX-RS can convert this string data into any Java type that you want, provided that it matches one of the following criteria:
> 1. It is a primitive type. The int, short, float, double, byte, char, and boolean types all fit into this category.
> 2. It is a Java class that has a constructor with a single String parameter.
> 3. **It is a Java class that has a static method named valueOf() that takes a single String argument and returns an instance of the class.**
> 4. It is a java.util.List<T>, java.util.Set<T>, or java.util.SortedSet<T>, where T is a type that satisfies criteria 2 or 3 or is a String. Examples are List<Double>, Set<String>, or SortedSet<Integer>.

https://dennis-xlc.gitbooks.io/restful-java-with-jax-rs-2-0-en/content/en/part1/chapter5/common_functionality.html

georgeracu:
Yep, as @heathd said above, we used `.of` in other value objects to map from Java primitives to the type we need. This is a pattern that is used in Java 8 in class `Optional.of()`. Unfortunately Java people were not consistent in usage of `valueOf` vs `of` as they use `valueOf` in wrapper objects like `Integer, Long etc.`. JAX-RS is using `valueOf` so we have both now.

danworth:
Ah thanks @heathd  & @georgeracu that makes sense.

PR #213, opened 2018-07-09
Pp 3680 add external ids

oswaldquek:
This PR adds to the direct debit event:
- an external id for the event itself
- the mandate external id
- the transaction external id

The mandate and transaction external ids are added because the publicapi is going to need them to create links for the payment (transaction) and direct debit (mandate) for searching for events (see https://paydd.docs.apiary.io/#reference/events/events/get-events).


src/main/java/uk/gov/pay/directdebit/events/api/DirectDebitEventsResponse.java
```diff
@@ -23,7 +24,9 @@
     private DirectDebitEventsPagination pagination;
     
     public DirectDebitEventsResponse(List<DirectDebitEvent> events, int page, int total, DirectDebitEventsPagination directDebitEventsPagination) {
-        this.events = events;
+        for (DirectDebitEvent event: events) {
+            this.events.add(new DirectDebitEventExternalView(event));
+        }
```
alexbishop1: Alternatively, you could remove the initialisation of the `events` field with a `new ArrayList<>()` and do this:

```java
this.events = events.stream().map(DirectDebitEventExternalView::new).collect(toList());
```

src/main/java/uk/gov/pay/directdebit/payments/dao/DirectDebitEventDao.java
```diff
@@ -1,26 +1,114 @@
...
+        if (nonNull(searchParams.getTransactionExternalId())) {
+            searchStrings.add("t.external_id = :transaction_id");
+            queryMap.put("transaction_id", searchParams.getTransactionExternalId());
+        }
+        if (nonNull(searchParams.getBeforeDate())) {
+            searchStrings.add("e.event_date < :before_date");
+            queryMap.put("before_date", searchParams.getBeforeDate());
+        }
+        if (nonNull(searchParams.getAfterDate())) {
+            searchStrings.add("e.event_date > :after_date");
```
alexbishop1: I think `>=` might be more consistent with what we do for other date-based searches

src/main/java/uk/gov/pay/directdebit/payments/dao/DirectDebitEventDao.java
```diff
@@ -1,26 +1,114 @@
...
+        return (long) jdbi.withHandle(handle -> {
+            return handle.createUpdate("INSERT INTO events(mandate_id, external_id, transaction_id, event_type, " +
+                    "event, event_date) VALUES (:mandateId, :externalId, :transactionId, :eventType, :event, :eventDate)")
+                    .bindBean(directDebitEvent)
+                    .execute();
+        });
+    }
+
+    public Optional<DirectDebitEvent> findByMandateIdAndEvent(Long mandateId, DirectDebitEvent.Type eventType, DirectDebitEvent.SupportedEvent event) {
+        return jdbi.withHandle(handle -> handle.createQuery("SELECT e.id, e.mandate_id, e.transaction_id, e.event_type, e.event, e.event_date, e.version, e.external_id, " +
```
alexbishop1: Can you not use the `QUERY` constant here?

src/main/java/uk/gov/pay/directdebit/payments/params/DirectDebitEventSearchParams.java
```diff
@@ -7,45 +7,46 @@
...
     @Getter private final Integer page;
 
     public DirectDebitEventSearchParamsBuilder copy() {
-        return new DirectDebitEventSearchParamsBuilder().page(page).pageSize(pageSize).transactionId(transactionId).mandateId(mandateId).afterDate(afterDate).beforeDate(beforeDate);
+        return new DirectDebitEventSearchParamsBuilder().page(page).pageSize(pageSize).transactionExternalId(transactionExternalId).mandateExternalId(mandateExternalId).afterDate(afterDate).beforeDate(beforeDate);
     }
 
     public Map<String, String> getParamsAsMap() {
-        Map<String, String> params = new HashMap<>();
+        Map<String, String> params = new LinkedHashMap<>();
```
alexbishop1: Why do we want to preserve the order here?

src/main/resources/migrations/00035_alter_table_events_add_external_id.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE events ADD COLUMN external_id VARCHAR(26);
```
alexbishop1: Due to the way our database transactions work (we have to explicitly initiate a migration: they don’t usually run automatically on start-up), we tend to prefer to have a separate PR with the database changes. Otherwise, the new code will be running alongside the old database schema and won’t work until the migration is done. Separating the migration into a prior PR avoids this problem.

src/test/java/uk/gov/pay/directdebit/events/api/DirectDebitEventsPaginationTest.java
```diff
@@ -38,43 +37,42 @@ public void testWithAllSearchParameters() {
                 .page(2)
                 .afterDate(ZonedDateTime.of(2018, 6, 29, 8, 0, 0, 0, ZoneId.of("UTC")))
                 .beforeDate(ZonedDateTime.of(2018, 6, 29, 9, 0, 0, 0, ZoneId.of("UTC")))
-                .mandateId(1234L)
-                .transactionId(5678L)
+                .mandateExternalId("1234L")
+                .transactionExternalId("5678L")
```
alexbishop1: These look like strings that used to be longs…

(It’s fine because they’re basically arbitrary.)

src/test/java/uk/gov/pay/directdebit/payments/dao/DirectDebitDirectDebitEventDaoIT.java
```diff
@@ -72,8 +73,9 @@ public void shouldInsertAnEvent() {
 
     @Test
     public void shouldFindByMandateIdAndEvent() {
-        aDirectDebitEventFixture()
+        DirectDebitEventFixture insert = aDirectDebitEventFixture()
```
alexbishop1: Maybe call this `directDebitEvent` or something rather than `insert`?

src/main/java/uk/gov/pay/directdebit/payments/dao/DirectDebitEventDao.java
oswaldquek: not really. they are subtly different. The constant `QUERY`'s searchFields have a dynamic WHERE clause, whereas the query here does not.

src/main/java/uk/gov/pay/directdebit/payments/params/DirectDebitEventSearchParams.java
oswaldquek: It's not really required from a production-code perspective, but it was to make the tests in DirectDebitEventsPaginationTest pass as we do something like
```
assertEquals(PaginationLink.ofValue(testUrl + "/test/?before=2018-06-29T09%3A00%3A00Z&after=2018-06-29T08%3A00%3A00Z&mandate_external_id=1234L&transaction_external_id=5678L&page=2&page_size=10")
                , pagination.getSelfLink());
```
Perhaps we should perform the assertion in a better way that makes the ordering irrelevant?

src/test/java/uk/gov/pay/directdebit/events/api/DirectDebitEventsPaginationTest.java
oswaldquek: yeah, the strings are arbitrary... as long as they're strings :)

src/main/resources/migrations/00035_alter_table_events_add_external_id.sql
oswaldquek: https://github.com/alphagov/pay-direct-debit-connector/pull/218

src/main/java/uk/gov/pay/directdebit/payments/params/DirectDebitEventSearchParams.java
alexbishop1: I thought it would be something to do with tests. I can see why you did this. Improving the assertion is a pain (you’d probably have to reparse the URL into a set or something) but could probably be reused everywhere we build up URLs like this. Keeping the order deterministic is also a reasonable solution. I’ll leave this one up to you.

oswaldquek: let's keep it deterministic for now then

PR #217, opened 2018-07-11
BAU: Make tests which use DB work with docker for mac

heathd:
Docker for mac is connected to over a unix socket rather than TCP. The postgres setup code
attempts to calculate the hostname of the postgres server. Rather than doing that, we can use
the docker class itself, which knows the hostname that it's running on. This means we can delete
a bunch of code.

## How to test

I've only tested this on docker-for-mac so I'd appreciate if someone could test it on a docker-toolbox setup as well.

To use with docker for mac, you need to ensure that there are no `DOCKER_*` env vars set in your run configuration, and also that intellij is not propagating the env vars through from its environment.

### Logging


heathd:
I've now tested this with docker toolbox as well

PR #221, opened 2018-07-13
BAU fix db script breakage

SandorArpa:
## WHAT YOU DID
- add 'if not exists' to liquibase script
- rename script to follow logical numbering

## How to test
- run `msl -l up` locally on branch. make sure that `external_id` in `events` exists
- logs should show that the script ran but no action were taken

### DIFF
--- a/src/main/resources/migrations/00037_alter_table_events_add_external_id.sql
+++ b/src/main/resources/migrations/00037_alter_table_events_add_external_id.sql
@@ -1,0 +1,7 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:alter_table_events_add_external_id
+ALTER TABLE events DROP COLUMN IF EXISTS external_id;
+ALTER TABLE events ADD COLUMN external_id VARCHAR(26);
+
+--rollback ALTER TABLE events DROP COLUMN external_id;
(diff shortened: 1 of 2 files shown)


src/main/resources/migrations/00037_alter_table_events_add_external_id.sql
```diff
@@ -0,0 +1,6 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:alter_table_events_add_external_id
+ALTER TABLE events ADD COLUMN IF NOT EXISTS external_id VARCHAR(26);
```
DanailMinchev: Needs upgrading PostgreSQL which supports this syntax

DanailMinchev: Currently we use `govukpay/postgres:9.4.4` in https://github.com/alphagov/pay-direct-debit-connector/blob/ec030adff4fa88fd027c70ca5b7d92e70799d995/src/test/java/uk/gov/pay/directdebit/junit/DropwizardAppWithPostgresRule.java#L30, but `IF NOT EXISTS` was added in `9.6` https://www.postgresql.org/docs/9.6/static/sql-altertable.html

We need to rebuild the image with the new version and push it to DockerHub or use 9.4.4 syntax

PR #223, opened 2018-07-13
PP-3646 Implement `state` search and fix pagination

DanailMinchev:
## WHAT YOU DID

- add sql generation for `state` using mapping from `ExternalPaymentState`
- handle external states from resource level downwards
- fix pagination (it's page and not page_number)
- fix bug in calculating next/last/prev page
- update docs
- add/update relevant tests
- fix pagination url (no & after ?)
- added unit tests for `ViewPaginationBuilder`
- refactor `ViewPaginationBuilder` (remove unnecessary list of PaymentViewResultResponse

with @SandorArpa

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/payments/model/ViewPaginationBuilder.java
+++ b/src/main/java/uk/gov/pay/directdebit/payments/model/ViewPaginationBuilder.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.directdebit.payments.model;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import uk.gov.pay.directdebit.payments.api.PaymentViewResultResponse;
 import uk.gov.pay.directdebit.payments.links.PaginationLink;
@@ -10,6 +11,9 @@ import javax.ws.rs.core.UriInfo;
 import java.net.URI;
 import java.util.List;
 
+import static com.fasterxml.jackson.annotation.JsonInclude.Include;
+
+@JsonInclude(Include.NON_NULL)
 public class ViewPaginationBuilder {
 
     private static final String SELF_LINK = "self";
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/pay/directdebit/payments/model/ViewPaginationBuilder.java
```diff
@@ -80,10 +84,13 @@ private void buildLinks(long lastPage) {
         lastLink = PaginationLink.ofValue(uriWithParams(searchParams.buildQueryParamString()).toString());
 
         searchParams.withPage(selfPageNum - 1);
-        prevLink = selfPageNum == 1L ? null : PaginationLink.ofValue(uriWithParams(searchParams.buildQueryParamString()).toString());;
+        prevLink = selfPageNum == 1L ? null :
```
simad: there is a fairly amount of logic here and I'd like to see it tested. I know it is implicitly tested elsewhere (like in `PaymentViewServiceTest.java`) but with unit tests here you could check all the paths without having to actually mess with fixtures and so on

src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
```diff
@@ -29,15 +39,32 @@
...
+    private String state;
     private SearchDateParams searchDateParams;
     private PaginationParams paginationParams;
     private Map<String, Object> queryMap;
-    
+
+    private static Map<String, String> externalPaymentToInternalStateQueryMap = new HashMap<>(4);
+
+    static {
+        String andStateCondition = " AND t.state = '%s'";
```
simad: I find this `andStateCondition` here not really readable. I would much rather put this in the StringBuilder below, see other comment.
The map would be simply: 

```
externalPaymentToInternalStateQueryMap
     .put(ExternalPaymentState.EXTERNAL_STARTED.getState(), NEW);
etc.
```

src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
```diff
@@ -137,6 +171,9 @@ public String generateQuery() {
         if (amount != null) {
             sb.append(" AND t.amount = :" + AMOUNT_FIELD);
         }
+        if (isNotBlank(state) && externalPaymentToInternalStateQueryMap.containsKey(state)) {
+            sb.append(externalPaymentToInternalStateQueryMap.get(state));
```
simad: here I would do: 
`sb.append(format(" AND t.state = '%s'", externalPaymentToInternalStateQueryMap.get(state)));`

I think it is clearer what's going on, and it is similar to the other fields

src/main/java/uk/gov/pay/directdebit/payments/params/PaymentViewSearchParams.java
DanailMinchev: Great suggestion! :) Will change that, not sure how to handle the case with https://github.com/alphagov/pay-direct-debit-connector/pull/223/files#diff-1f282af1a2bd8e2a97f7d49673214372R60 ?

simad: We have clarified this via message.

src/main/java/uk/gov/pay/directdebit/payments/model/ViewPaginationBuilder.java
DanailMinchev: Added in https://github.com/alphagov/pay-direct-debit-connector/pull/223/commits/c53d425f52adf6d6c6b3b223d5ce71bb99ab5760

PR #225, opened 2018-07-13
PP-3835 Refactor of services

simad:
## WHAT YOU DID
This work takes heavy inspiration from @mrlumbu original exploration around the refactoring of services. We were doing some mildly overcooked spaghetti before, going straight from resource-level to a provider specific (sandbox/gocardless) service, which was delegating to a mandate service with a not so nice back and forth.

This PR aims to make the logic a bit more straightforward and uni-directional:
- Introduced two different mandate services (one-off and on-demand) which take care of acting on a specific type of mandate. The only common operation between the two is currently confirming a mandate. All the others are quite specific (ie. collect only works for a on demand mandate). 
The original work had that in the `MandateCommandService` interface, so that both services were overriding the code. I liked that approach, but I've realised it over-complicates things a bit because in most cases we already know which kind of mandate we are operating on, so there's no real need to use a factory. 
An example of this is the create payment endpoint - we already know it's going to create a one-off mandate, so we can use that service directly.
- Split mandate service into query service and update state service. This operates under the assumption that it's better to have more small-ish services rather than one single `god` service which does everything. This pattern can be applied to transaction service, too, moving forward.
- Simplified SandboxService to the core, it does basic logging and returns some mock charge date 
- Removed the confirm mandate resource with a single /confirm endpoint and moved it in the mandate resource
- Mapped most of `Map<String, String>` representing requests coming in endpoints to domain objects. This helped getting rid of parsers.
- Shuffled tests around


src/main/java/uk/gov/pay/directdebit/mandate/api/ConfirmMandateRequest.java
```diff
@@ -0,0 +1,42 @@
...
+    
+    @JsonProperty("account_number")
+    private String accountNumber;
+
+    @JsonProperty("transaction_external_id")
+    private String transactionExternalId;
+
+    public ConfirmMandateRequest(String sortCode, String accountNumber, String transactionExternalId) {
+        this.sortCode = sortCode;
+        this.accountNumber = accountNumber;
```
mrlumbu: these should be of type `SortCoe` and AccountNumber` now I think?

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateCommandService.java
```diff
@@ -0,0 +1,9 @@
+package uk.gov.pay.directdebit.mandate.services;
+
+import uk.gov.pay.directdebit.gatewayaccounts.model.GatewayAccount;
+import uk.gov.pay.directdebit.mandate.api.ConfirmMandateRequest;
+import uk.gov.pay.directdebit.mandate.model.Mandate;
+
+public interface MandateCommandService {
+    void confirm(GatewayAccount gatewayAccount, Mandate mandate, ConfirmMandateRequest confirmDetailsRequest);
```
mrlumbu: could add `create` and potentially `collect` to this interface

src/main/java/uk/gov/pay/directdebit/mandate/services/OnDemandMandateService.java
```diff
@@ -0,0 +1,84 @@
...
+            TransactionService transactionService,
+            MandateService mandateService) {
+        this.paymentProviderFactory = paymentProviderFactory;
+        this.mandateStateUpdateService = mandateStateUpdateService;
+        this.transactionService = transactionService;
+        this.mandateService = mandateService;
+    }
+
+    @Override
+    public void confirm(
```
mrlumbu: I think it would be nice if we had some uniformity about what service methods returned - i.e. probably the updated object? But no biggie

src/main/java/uk/gov/pay/directdebit/mandate/services/OneOffMandateService.java
```diff
@@ -0,0 +1,58 @@
...
+            TransactionService transactionService,
+            MandateStateUpdateService mandateStateUpdateService,
+            MandateService mandateService){
+        this.paymentProviderFactory = paymentProviderFactory;
+        this.transactionService = transactionService;
+        this.mandateStateUpdateService = mandateStateUpdateService;
+        this.mandateService = mandateService;
+    }
+    
+    public Transaction create(GatewayAccount gatewayAccount,
```
mrlumbu: I find it strange that we create a mandate but return a transaction

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateStateUpdateService.java
```diff
@@ -0,0 +1,123 @@
...
+        mandateDao.updateState(mandate.getId(), newState);
+        LOGGER.info("Updating mandate {} - from {} to {}",
+                mandate.getExternalId(),
+                mandate.getState(),
+                newState);
+        mandate.setState(newState);
+        return mandate;
+    }
+    
+    void canUpdateStateFor(Mandate mandate, DirectDebitEvent.SupportedEvent event) {
```
mrlumbu: I know I wrote this method, but it is a bit weird - I would prefer it to return a Boolean rather than throwing

src/main/java/uk/gov/pay/directdebit/mandate/api/ConfirmMandateRequest.java
simad: Yes, good catch. I think we used account number and sortcode everywhere except here. I originally thought this would be an object which get serialised (hence the `@jsonproperty`) annotations, but it's actually just a domain object, so no need to have properties as strings

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateStateUpdateService.java
mrlumbu: but again, not a big deal

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateCommandService.java
simad: We had it like that, but then I've realised (see description of the PR) that it adds quite a bit of complexity, because create and collect are used only for specific kind of mandates.
- collect is used only for on demand
- create is used only for one-off in the TransactionResource
- create is used only for on demand in the create agreement endpoint 

So, instead of getting the factory and switching on mandate type, I've used the specific mandate service right there, and that makes adding create and collect to the common interface pointless (for now)

src/main/java/uk/gov/pay/directdebit/mandate/services/OnDemandMandateService.java
simad: yes, that would've been cool, will se if I can make it work but probably will keep it like this for now :D the PR is quite big

src/main/java/uk/gov/pay/directdebit/mandate/services/OneOffMandateService.java
simad: true, could return both but mandate is not needed upstream, so a bit of a weird one

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateStateUpdateService.java
simad: hahaha...not sure about that, i think it fits well with the exception mapper pattern we are using. It's a bit of a guard. If it returns a boolean then we just move the logic (if it can't update, throw) one level above

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateCommandService.java
mrlumbu: fair enough. I can def see your point with `collect`, but `create` seems naturally part of the interface, even if doesn't need to be from the point of view of how you've used it. But now I'm just imposing my own prejudices

src/main/java/uk/gov/pay/directdebit/mandate/services/OneOffMandateService.java
mrlumbu: well, we currently only need the transaction, but still seems more natural for it to return the thing it created, then if the calling context needs the transaction, it can get it from the mandate. But fine with it as is

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateStateUpdateService.java
mrlumbu: jolly good

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateCommandService.java
simad: I think you're right, as in collect definitely feels wrong to have it in the interface (because it's not common), whereas create could be there even if it's not really used that way. I'll add it.

mrlumbu: sure

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateServiceFactory.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public OnDemandMandateService getOnDemandMandateService() {
+        return onDemandMandateService;
+    }
+
+    public OneOffMandateService getOneOffMandateService() {
+        return oneOffMandateService;
+    }
+
+    public MandateCommandService getMandateControlService(MandateType mandateType) {
```
danworth: We seem to mix Command and Control terminology (`MandateCommandService`  whilst method is named `getMandateControlService`). Should there be a consistent use?

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/GoCardlessPaymentHandlerTest.java
```diff
@@ -37,7 +37,7 @@
     @Mock
     private PayerService mockedPayerService;
     @Mock
-    private GoCardlessService mockedGoCardlessService;
+    private GoCardlessEventService mockedGoCardlessService;
```
georgeracu: Maybe `mockedGoCardlessEventService`. Minor point

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessServiceTest.java
```diff
@@ -37,7 +28,7 @@
 
 public class WebhookGoCardlessServiceTest {
     @Mock
-    private GoCardlessService mockedGoCardlessService;
+    private GoCardlessEventService mockedGoCardlessService;
```
georgeracu: Maybe mockedGoCardlessEventService. Minor point

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/GoCardlessHandlerTest.java
```diff
@@ -29,7 +29,7 @@
     TransactionService mockedTransactionService;
 
     @Mock
-    GoCardlessService mockedGoCardlessService;
+    GoCardlessEventService mockedGoCardlessService;
```
georgeracu: Maybe mockedGoCardlessEventService. Minor point

src/test/java/uk/gov/pay/directdebit/mandate/resources/MandateResourceIT.java
```diff
@@ -387,6 +425,235 @@ public void shouldChangePaymentType_ifMandateIsOneOff() {
...
+                .withGatewayAccountFixture(gatewayAccountFixture)
+                .withPayerFixture(payerFixture)
+                .insert(testContext.getJdbi());
+
+        TransactionFixture transactionFixture = aTransactionFixture()
+                .withMandateFixture(mandateFixture)
+                .withState(PaymentState.NEW)
+                .insert(testContext.getJdbi());
+
+        String lastTwoDigitsBankAccount = payerFixture.getAccountNumber().substring(payerFixture.getAccountNumber().length() - 2);
```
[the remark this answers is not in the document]
simad: Great idea!

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/GoCardlessHandlerTest.java
simad: good catch

PR #231, opened 2018-07-19
PP-3838 Expose internal mandate state to frontend

simad:
## WHAT YOU DID
We need the internal state of a mandate to properly deal with error messages in frontend (ie. know if a mandate has expired)


georgeracu:
I might be late at the party, but shouldn't we use `external states` to map to internal states? By sending internal `connector` specific states all the way to front-end then we couple the two microservices on internals of `connector`, we are not consistent on what we do (use external_state to send states to front-end from connector) and we have something not working in our code if we need internal states to display error messages. I know we are very pressured on time but we should strive more to be consistent in the way we do things.

PR #242, opened 2018-07-26
PP-4048 Make sure transaction has the updated mandate reference

simad:
## WHAT YOU DID
We are modifying the mandate reference, but the mandate embedded in transaction was not updated correctly for one-off mandates.

### DIFF
--- a/src/test/java/uk/gov/pay/directdebit/mandate/resources/MandateResourceIT.java
+++ b/src/test/java/uk/gov/pay/directdebit/mandate/resources/MandateResourceIT.java
@@ -533,7 +533,8 @@ public class MandateResourceIT {
     @Test
     public void confirm_shouldCreateACustomerBankAccountMandateAndUpdateCharge_ForGoCardless_ifMandateIsOneOff() {
         GatewayAccountFixture gatewayAccountFixture = GatewayAccountFixture.aGatewayAccountFixture()
-                .withPaymentProvider(GOCARDLESS).insert(testContext.getJdbi());
+                .withPaymentProvider(GOCARDLESS)
+                .insert(testContext.getJdbi());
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/pay/directdebit/mandate/resources/MandateResourceIT.java
```diff
@@ -566,9 +567,9 @@ public void confirm_shouldCreateACustomerBankAccountMandateAndUpdateCharge_ForGo
                 "  \"template\": \"ONE_OFF_PAYMENT_CONFIRMED\",\n" +
                 "  \"personalisation\": {\n" +
                 "    \"amount\": \"" + BigDecimal.valueOf(transactionFixture.getAmount(), 2).toString() + "\",\n" +
-                "    \"mandate reference\": \"" + mandateFixture.getMandateReference() + "\",\n" +
+                "    \"mandate reference\": \"REF-123\",\n" +
                 "    \"bank account last 2 digits\": \"" + lastTwoDigitsBankAccount + "\",\n" +
-                "    \"collection date\": \"2014-05-21\",\n" +
+                "    \"collection date\": \"21/05/2014\",\n" +
```
georgeracu: Just a small question: have we changed the date format?

DanailMinchev: Looks like we are using:
```
LocalDate.now().plusDays(4).format(DateTimeFormatter.ofPattern("dd/MM/yyyy"))
```
in other tests as format

simad: Yes, It was just a problem of this test. Unfortunately if the email is not matched the test passes, even though the mismatch is reported in the console

PR #248, opened 2018-07-27
PP-4002 Update Mandate to hold creditorId

DanailMinchev:
## WHAT YOU DID

Update `GoCardlessMandate` to hold `gocardlessCreditorId`

- Added `GoCardlessCreditorId` type.
- Updated `GoCardlessMandate` to add `gocardlessCreditorId`.
- Updated associated fixtures, DAOs and tests.

with @danworth


src/main/java/uk/gov/pay/directdebit/common/model/subtype/CreditorId.java
```diff
@@ -0,0 +1,38 @@
...
+    public static CreditorId valueOf(String creditorId) {
+        return CreditorId.of(creditorId);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        CreditorId that = (CreditorId) o;
+        return Objects.equals(creditorId, that.creditorId);
```
simad: I think we are using the intellij default for equal and hashcode everywhere else. I am fine using the `Objects` one, though, it is quite compact.

src/test/java/uk/gov/pay/directdebit/payments/clients/GoCardlessClientFacadeTest.java
```diff
@@ -100,4 +113,23 @@ public void validateReturnsGoCardlessBankAccountLookupWithNullBankNameAndBacsFal
...
+                MandateFixture.aMandateFixture().toEntity();
+        GoCardlessCustomer goCardlessCustomer =
+                GoCardlessCustomerFixture.aGoCardlessCustomerFixture().toEntity();
+
+        given(mockMandate.getId()).willReturn("test-mandate-id-here");
+        given(mockMandate.getReference()).willReturn("test-mandate-reference-here");
+        given(mockMandate.getLinks()).willReturn(mockMandateLinks);
+        given(mockMandateLinks.getCreditor()).willReturn(creditorId.toString());
+
+        given(mockGoCardlessClientWrapper.createMandate(any(), any())).willReturn(mockMandate);
```
simad: I really discourage using `any()`, there are very few cases where we don't know what the parameters are going to be, especially when we actually set up the mocks ourselves. We have the mandate external Id and also the goCardlessCustomer, let's use them

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
```diff
@@ -55,6 +59,7 @@
             "  m.type AS mandate_type," +
             "  m.state AS mandate_state," +
             "  m.created_date AS mandate_created_date," +
+            "  m.creditor_id AS mandate_creditor_id," +
```
simad: As discussed with @DanailMinchev, we have the concept of `Mandate` which is purely internal to our system, and the concept of `GoCardlessMandate` which is the 'equivalent' mandate in gocardless. As `CreditorId` is 100% gocardless specific (it is an identifier used for one of their resources) - we should keep these models separate. `CreditorId` should be part of `GoCardlessMandate`, as, for example,`goCardlessMandateId`.

We can have a broader discussion on Monday if needed.

src/main/java/uk/gov/pay/directdebit/common/model/subtype/CreditorId.java
DanailMinchev: Changed it to `IntelliJ IDEA` style for consistency

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
DanailMinchev: Thank you, well spotted!
Created PR for database migrations - https://github.com/alphagov/pay-direct-debit-connector/pull/249
After #249 is merged we can address this comment.

src/main/java/uk/gov/pay/directdebit/mandate/dao/mapper/GoCardlessMandateMapper.java
```diff
@@ -1,22 +1,26 @@
...
 
     @Override
     public GoCardlessMandate map(ResultSet resultSet, StatementContext statementContext) throws SQLException {
         return new GoCardlessMandate(
                 resultSet.getLong(ID_COLUMN),
                 resultSet.getLong(MANDATE_ID_COLUMN),
                 resultSet.getString(GOCARDLESS_MANDATE_ID_COLUMN),
-                null);
+                null,
+                GoCardlessCreditorId.of(resultSet.getString(GOCARDLESS_CREDITOR_ID)));
```
alexbishop1: Is this safe to do? Existing GoCardless mandates in our database aren’t going to have a `gocardless_creditor_id` value, which means you’ll be passing `null` to the `GoCardlessCreditorId.of(String)`, which will go _kaboom!_ because it doesn’t accept `null` parameters.

If you look at `GatewayAccountMapper`, you can see how we work around this for other entities (tl;dr: with some horrible `if` statements).

src/main/java/uk/gov/pay/directdebit/mandate/model/GoCardlessMandate.java
```diff
@@ -1,20 +1,21 @@
...
+    private final GoCardlessCreditorId goCardlessCreditorId;
 
-    public GoCardlessMandate(Long id, Long mandateId, String goCardlessMandateId, String goCardlessReference) {
+    public GoCardlessMandate(Long id, Long mandateId, String goCardlessMandateId, String goCardlessReference, GoCardlessCreditorId goCardlessCreditorId) {
         this.id = id;
         this.mandateId = mandateId;
         this.goCardlessMandateId = goCardlessMandateId;
         this.goCardlessReference = goCardlessReference;
-    }
-    public GoCardlessMandate(Long mandateId, String goCardlessMandateId, String goCardlessReference) {
```
alexbishop1: Ah, I see, you did remove this constructor.

I actually quite like constructors like this because it makes it clear that the `id` can legitimately be `null`.

src/main/java/uk/gov/pay/directdebit/mandate/model/GoCardlessMandate.java
```diff
@@ -37,6 +38,10 @@ public String getGoCardlessReference() {
         return goCardlessReference;
     }
 
+    public GoCardlessCreditorId getGoCardlessCreditorId() {
```
alexbishop1: If this could be `null` (and I think legacy GoCardless mandates in the database mean it might be), we may wish to return an `Optional<GoCardlessCreditorId>` here.

src/test/java/uk/gov/pay/directdebit/common/clients/GoCardlessClientFacadeTest.java
```diff
@@ -100,4 +112,23 @@ public void validateReturnsGoCardlessBankAccountLookupWithNullBankNameAndBacsFal
         assertThat(result.isBacs(), is(false));
     }
 
+    @Test
+    public void createMandateReturnsGoCardlessMandateWithCreditorId() {
```
alexbishop1: Since this is the only test that tests the `GoCardlessClientWrapper.createMandate(MandateExternalId, GoCardlessCustomer)` method, we may as well want to make it a more general test and have it assert other stuff we know about the returned mandate (e.g. the IDs and reference).

src/test/java/uk/gov/pay/directdebit/payments/services/GoCardlessServiceTest.java
```diff
@@ -49,6 +50,7 @@
     static final String TRANSACTION_ID = "sdkfhsd2jfhjdks";
     static final SortCode SORT_CODE = SortCode.of("123456");
     static final AccountNumber ACCOUNT_NUMBER = AccountNumber.of("12345678");
+    static final GoCardlessCreditorId CREDITOR_ID = GoCardlessCreditorId.of("gocardless-test-creditor-id-here");
```
alexbishop1: This doesn’t actually ever seem to be used

src/main/java/uk/gov/pay/directdebit/mandate/model/GoCardlessMandate.java
DanailMinchev: As discussed if we have more than two constructors we should use Builder pattern and we are going to rollback this constructor.

src/main/java/uk/gov/pay/directdebit/common/clients/GoCardlessClientFacade.java
```diff
@@ -48,9 +49,11 @@ public GoCardlessCustomer createCustomerBankAccount(MandateExternalId mandateExt
     public GoCardlessMandate createMandate(Mandate mandate, GoCardlessCustomer customer) {
         com.gocardless.resources.Mandate gcMandate = goCardlessClientWrapper.createMandate(mandate.getExternalId(), customer);
         return new GoCardlessMandate(
+                null,
```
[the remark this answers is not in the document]
DanailMinchev: As discussed if we have more than two constructors we should use Builder pattern and we are going to rollback this constructor.

src/main/java/uk/gov/pay/directdebit/mandate/model/GoCardlessMandate.java
DanailMinchev: As discussed this cannot be null in real case, but only in our existing GC mandates. We are going to create another DB migration to initialise the existing creditor id with dummy value and then remove it in this PR

src/main/java/uk/gov/pay/directdebit/mandate/dao/mapper/GoCardlessMandateMapper.java
DanailMinchev: As discussed it is now safe because we merged https://github.com/alphagov/pay-direct-debit-connector/pull/253

src/test/java/uk/gov/pay/directdebit/payments/services/GoCardlessServiceTest.java
DanailMinchev: Nice spot! We forgot to remove it after renaming.

PR #251, opened 2018-07-30
PP-3835 Add MANDATE_CREATED event

simad:
## WHAT YOU DID
 - We should record an event when we create and store
   a mandate in our DB. Otherwise, for example, on demand
   mandates will have 0 events until the token exchange
## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging


src/test/java/uk/gov/pay/directdebit/payments/services/DirectDebitEventServiceTest.java
```diff
@@ -224,6 +225,19 @@ public void registerMandatePendingEventFor_shouldCreateExpectedEvent() {
         assertThat(directDebitEvent.getEventDate(), is(ZonedDateTimeMatchers.within(10, ChronoUnit.SECONDS, ZonedDateTime.now())));
     }
 
+    @Test
+    public void registerMandateCreatedEventFor_shouldCreateExpectedEvent() {
+        service.registerMandateCreatedEventFor(mandateFixture.toEntity());
+
+        verify(mockedDirectDebitEventDao).insert(prCaptor.capture());
```
alexbishop1: I assume `prCaptor` is left over from when we had “payment requests”…

simad: organic growth by copy paste

PR #255, opened 2018-08-01
PP-4043 Link GoCardless account to GOV.UK Pay account

georgeracu:
## WHAT YOU DID
- Added a table that will hold state values used to send to GoCardless
- The state value is returned to us from GoCardless and we will use it
to check if is valid against a gateway account external id

## How to test

- check if migration run successfully

### Logging

- [x] only emit log lines at ERROR level which require immediate attention from a support engineer. These will trigger a zendesk alert.

### DIFF
--- a/src/main/resources/migrations/00043_create_table_oauth_states.sql
+++ b/src/main/resources/migrations/00043_create_table_oauth_states.sql
@@ -3,7 +3,7 @@
 --changeset uk.gov.pay:add_table-gocardless_partner_app_account_connect_tokens
 CREATE TABLE gocardless_partner_app_account_connect_tokens (
   id BIGSERIAL PRIMARY KEY,
-  gateway_account_id BIGSERIAL NOT NULL,
+  gateway_account_id BIGINT NOT NULL,
   token VARCHAR(26) NOT NULL,
   active BOOLEAN DEFAULT TRUE
 );
--- a/src/main/resources/migrations/00043_create_table_oauth_states.sql
+++ b/src/main/resources/migrations/00043_create_table_oauth_states.sql
@@ -1,0 +1,15 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_table-oauth_states
+CREATE TABLE oauth_states (
+  id BIGSERIAL PRIMARY KEY,
+  gateway_account_external_id VARCHAR(26) NOT NULL,
+  state VARCHAR(26) NOT NULL,
+  active BOOLEAN DEFAULT TRUE,
+  version INTEGER DEFAULT 0 NOT NULL
+);
+--rollback drop table oauth_states;
+
+--changeset uk.gov.pay:add_oauth_states_gateway_accounts_fk
+ALTER TABLE oauth_states ADD CONSTRAINT oauth_states_gateway_accounts_fk FOREIGN KEY (gateway_account_external_id) REFERENCES gateway_accounts (external_id);
+--rollback drop constraint oauth_states_gateway_accounts_fk;
--- a/src/main/resources/migrations/00043_create_table_oauth_states.sql
+++ b/src/main/resources/migrations/00043_create_table_oauth_states.sql
@@ -1,15 +1,14 @@
 --liquibase formatted sql
 
---changeset uk.gov.pay:add_table-oauth_states
-CREATE TABLE oauth_states (
+--changeset uk.gov.pay:add_table-gocardless_partner_app_account_connect_tokens
+CREATE TABLE gocardless_partner_app_account_connect_tokens (
   id BIGSERIAL PRIMARY KEY,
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/00043_create_table_oauth_states.sql
```diff
@@ -0,0 +1,15 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_table-oauth_states
+CREATE TABLE oauth_states (
+  id BIGSERIAL PRIMARY KEY,
+  gateway_account_external_id VARCHAR(26) NOT NULL,
+  state VARCHAR(26) NOT NULL,
```
DanailMinchev: Is `26` chars enough for the state? Could you give an example of the values? Thanks

src/main/resources/migrations/00043_create_table_oauth_states.sql
```diff
@@ -0,0 +1,15 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_table-oauth_states
+CREATE TABLE oauth_states (
```
alexbishop1: Can we come up with a more descriptive name for this?

I understand these are effectively tokens or identifiers we use during the partner connect flow so something that tries to capture that?

OAuth is just the mechanism used for connecting but doesn’t really matter from the database perspective and “state” can mean many things (I assume its name comes from the name of the parameter GoCardless use).

src/main/resources/migrations/00043_create_table_oauth_states.sql
```diff
@@ -0,0 +1,15 @@
...
+  id BIGSERIAL PRIMARY KEY,
+  gateway_account_external_id VARCHAR(26) NOT NULL,
```
alexbishop1: If we refer to another table within the same database, we generally use the database ID (usually the `id` column), not the external ID (which we use if we’re referring to an entity in another database)

src/main/resources/migrations/00043_create_table_oauth_states.sql
georgeracu: The state is some value that we generate and I gave it 26 chars so we can reuse the external id generator to populate it :)

src/main/resources/migrations/00043_create_table_oauth_states.sql
georgeracu: I used `oauth_tokens`. I know it doesn't matter for the db but I think it matters to the developer who is going to look at the name of the table and try to figure out what is doing.
I believe `oauth_tokens` is misleading because this might confuse people in thinking that this token is the one we get from GoCardless, which is not true. 
This table unfortunately is again very specific to GoCardless so we might as well call it `gocardless_oauth_request_tokens`.

Any naming suggestions are more than welcome :)

src/main/resources/migrations/00043_create_table_oauth_states.sql
georgeracu: Changed it to reference `id` in `gateway_account`, thanks for suggesting it

src/main/resources/migrations/00043_create_table_oauth_states.sql
alexbishop1: It’s already very specific to GoCardless, so a GoCardless-specific name is fine. Call it `gocardless_partner_app_account_connect_tokens` something (`tokens` is unfortunately a very overloaded term: GoCardless’s own docs recommend the `state` parameter should have a “CSRF token” in it).

This table contains stuff related to connecting GoCardless accounts to our partner app (GoCardless unfortunately call this process “Connecting your users’ accounts” — “users” is such an overloaded term!) so I think it’s a good idea to try to capture in the name that it’s related to that process (`gocardless_partner_app_account_connect` is the best I can come up with right now).

Not keen on having `oauth` in the name: OAuth is just the mechanism used for the partner app account connect stuff. When we connect accounts, we’re not doing OAuth for the sake of OAuth, we’re doing it in order to connect accounts to our partner app. It’s conceivable we may end up with other OAuth functionality in our app in the future to do other things, which might mean a table name like `oauth_tokens` could cause confusion (also OAuth already has a concept of tokens, which could lead to more confusion).

src/main/resources/migrations/00043_create_table_oauth_states.sql
```diff
@@ -0,0 +1,14 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_table-gocardless_partner_app_account_connect_tokens
+CREATE TABLE gocardless_partner_app_account_connect_tokens (
+  id BIGSERIAL PRIMARY KEY,
+  gateway_account_id BIGSERIAL NOT NULL,
```
alexbishop1: I think we just want `BIGINT` here: we’ll be specifying the value every time rather than relying on a sequence providing us with one

PR #261, opened 2018-08-07
PP-4043 Drop UNIQUE constraint gateway_accounts_organisation_key

georgeracu:
- This is necessary as when updating column `organisation` in table
`gateway_accounts` it fails for the same value. Access token can change,
but the organisation will be the same when a new token is issued for an
existing merchant app.

### DIFF
--- a/src/main/resources/migrations/00045_drop_constraint_gateway_accounts_organisation_id.sql
+++ b/src/main/resources/migrations/00045_drop_constraint_gateway_accounts_organisation_id.sql
@@ -1,0 +1,5 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:alter_table_gateway_accounts_drop_constraint_gateway_accounts_organisation_key
+ALTER TABLE gateway_accounts DROP CONSTRAINT IF EXISTS gateway_accounts_organisation_key;
+--rollback ALTER TABLE gateway_accounts ADD UNIQUE (organisation);


src/main/resources/migrations/00045_drop_constraint_gateway_accounts_organisation_id.sql
```diff
@@ -0,0 +1,5 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:alter_table_gateway_accounts_drop_constraint_gateway_accounts_organisation_key
+ALTER TABLE gateway_accounts DROP CONSTRAINT IF EXISTS gateway_accounts_organisation_key;
```
danworth: Exploring in my local environment I seem to be able to update the access_token even with the unique constraint on organisation.

```
 id |               external_id               | payment_provider | service_name | type | description | analytics_id | version | access_token | organisation 
----+-----------------------------------------+------------------+--------------+------+-------------+--------------+---------+--------------+--------------
  2 | DIRECT_DEBIT:q7k1jdodep9qrnegc1bn6iifhg | SANDBOX          | My service   | TEST |             |              |       0 |              | 
  1 | DIRECT_DEBIT:3gnuulqn1vkfa2aiuf86q0vlqi | SANDBOX          | My service   | TEST |             |              |       0 | myToken      | myorg
```

Then updating the acces_token:

```
directdebit_connector=# UPDATE gateway_accounts SET access_token = 'myToke32' WHERE id=1;
UPDATE 1
directdebit_connector=# select * from gateway_accounts;
 id |               external_id               | payment_provider | service_name | type | description | analytics_id | version | access_token | organisation 
----+-----------------------------------------+------------------+--------------+------+-------------+--------------+---------+--------------+--------------
  2 | DIRECT_DEBIT:q7k1jdodep9qrnegc1bn6iifhg | SANDBOX          | My service   | TEST |             |              |       0 |              | 
  1 | DIRECT_DEBIT:3gnuulqn1vkfa2aiuf86q0vlqi | SANDBOX          | My service   | TEST |             |              |       0 | myToke32     | myorg
```

Not sure if its worth seeing checking how this works in the DAO before removing the constraint?

PR #307, opened 2018-11-20
PP-4268 Use InetSocketAddress.createUnresolved instead of setting default ProxySelector

DanailMinchev:
## WHAT YOU DID

This is small improvement to replace ProxySelector with InetSocketAddress.createUnresolved.

Currently we use custom ProxySelector ( https://github.com/alphagov/pay-direct-debit-connector/blob/6f423b5dddddf2039aa1f878c4865eefe2e226ae/src/main/java/uk/gov/pay/directdebit/common/proxy/CustomInetSocketAddressProxySelector.java ) which is used as a default proxy selector for any outbound Java requests https://github.com/alphagov/pay-direct-debit-connector/blob/6f423b5dddddf2039aa1f878c4865eefe2e226ae/src/main/java/uk/gov/pay/directdebit/DirectDebitConnectorApp.java#L93

We want to change the configuration and implementation and not to use custom ProxySelector when initialising GoCardless SDK

We should use `InetSocketAddress.createUnresolved(configuration.getProxyConfig().getHost(), configuration.getProxyConfig().getPort())` instead of `new InetSocketAddress(configuration.getProxyConfig().getHost(), configuration.getProxyConfig().getPort())`, because `new InetSocketAddress(...)` will resolve to IP address straight away and this IP address may be invalid in the future (we should use hostnames for our proxy address).

`InetSocketAddress.createUnresolved` will not resolve reverse proxy domain name to IP address, hence it will always be valid (even in case when the hostname changes it's IP address)

Removed the `ProxySelector.setDefault(customInetSocketAddressProxySelector)` call to apply `InetSocketAddress.createUnresolved`

related PR: https://github.com/alphagov/pay-direct-debit-connector/pull/271

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/DirectDebitConnectorApp.java
+++ b/src/main/java/uk/gov/pay/directdebit/DirectDebitConnectorApp.java
@@ -83,15 +83,16 @@ public class DirectDebitConnectorApp extends Application<DirectDebitConfig> {
 
     @Override
     public void run(DirectDebitConfig configuration, Environment environment) {
-        if (!configuration.getProxyConfig().getHost().isEmpty()) {
-            CustomInetSocketAddressProxySelector customInetSocketAddressProxySelector =
-                    new CustomInetSocketAddressProxySelector(
-                            ProxySelector.getDefault(),
-                            configuration.getProxyConfig().getHost(),
-                            configuration.getProxyConfig().getPort()
-                    );
-            ProxySelector.setDefault(customInetSocketAddressProxySelector);
-        }
+        // TODO: remove CustomInetSocketAddressProxySelector and it's usage if InetSocketAddress.createUnresolved works
+//        if (!configuration.getProxyConfig().getHost().isEmpty()) {
+//            CustomInetSocketAddressProxySelector customInetSocketAddressProxySelector =
+//                    new CustomInetSocketAddressProxySelector(
+//                            ProxySelector.getDefault(),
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/directdebit/DirectDebitConnectorApp.java
```diff
@@ -83,15 +83,16 @@ public DataSourceFactory getDataSourceFactory(DirectDebitConfig configuration) {
...
-        if (!configuration.getProxyConfig().getHost().isEmpty()) {
-            CustomInetSocketAddressProxySelector customInetSocketAddressProxySelector =
-                    new CustomInetSocketAddressProxySelector(
-                            ProxySelector.getDefault(),
-                            configuration.getProxyConfig().getHost(),
-                            configuration.getProxyConfig().getPort()
-                    );
-            ProxySelector.setDefault(customInetSocketAddressProxySelector);
-        }
+        // TODO: remove CustomInetSocketAddressProxySelector and it's usage if InetSocketAddress.createUnresolved works
```
simad: We should delete this comment if it works, but ok for now :D

heathd:
can you give a bit more background about the reasons for doing this?

DanailMinchev:
@heathd I have updated the commit message and PR description with more information.
In a nutshell it's a small improvement not to use ProxySelectors

Thanks

heathd:
👍

PR #337, opened 2019-01-07
PP-4584 Use custom proxy selector for outbound HTTPS requests

DanailMinchev:
## WHAT

- Add `CustomInetSocketAddressProxySelector` which sets proxy for all
  outbound HTTPS requests and uses new system properties introduced in
  #333
- `CustomInetSocketAddressProxySelector` will be used with GoCardless SDK
  when there is no explicit proxy setting on the SDK client

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/common/proxy/CustomInetSocketAddressProxySelector.java
+++ b/src/main/java/uk/gov/pay/directdebit/common/proxy/CustomInetSocketAddressProxySelector.java
@@ -1,0 +1,88 @@
+package uk.gov.pay.directdebit.common.proxy;
+
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.net.Proxy;
+import java.net.ProxySelector;
+import java.net.SocketAddress;
+import java.net.URI;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Custom {@link java.net.ProxySelector ProxySelector} which will use only one proxy and will return new instance each time.
+ *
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/directdebit/common/proxy/CustomInetSocketAddressProxySelector.java
```diff
@@ -0,0 +1,88 @@
...
+ * Custom {@link java.net.ProxySelector ProxySelector} which will use only one proxy and will return new instance each time.
+ *
+ * @see <a href="https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html">Java Networking and Proxies</a>
+ */
+public class CustomInetSocketAddressProxySelector extends ProxySelector {
+
+    /**
+     * Keep a reference on the previous default
+     */
+    private ProxySelector defaultProxySelector;
```
rhowe-gds: Would it be really bad form to make this an Optional? It'd make all the comparisons to null neater

DanailMinchev: I think it's good idea not to check for `null`s and use `Optional`s, but we do need the `defaultProxySelector` actually as default fallback, that is why it is passed in `constructor` as mandatory field, so it's not optional
I do check for `null` just in case as it is not clear to me when the security manager will not be null (from the logic of `java.net.ProxySelector#getDefault` method)
Let me know if you have suggestions, thanks

PR #348, opened 2019-01-29
Pp 4597 output timestamps in miliseconds

georgeracu:
## WHAT YOU DID
* Enforced timestamps to have millisecond precision in response to API calls
* Replaced String with ZonedDateTime as field type in API response objects
* Added custom JSON serializer that uses common formatter for millisecond precision
* Code cleanup


src/test/java/uk/gov/pay/directdebit/payments/model/CustomDateSerializerTest.java
```diff
@@ -0,0 +1,38 @@
...
+    private CustomDateSerializer serializer;
+
+    @Before
+    public void before() {
+        serializer = new CustomDateSerializer();
+    }
+
+    @Test
+    public void shouldSerializeWithMillisecondPrecision() throws IOException {
+        ZonedDateTime testValue = ZonedDateTime.parse("2019-01-29T11:34:53.849Z");
```
alexbishop1: I think this would be a better test if the input date had greater than millisecond precision (e.g. 2019-01-29T11:34:53.849630917Z) but the final assertion checked that only millisecond precision was output

alexbishop1:
Correction — I said: “It won’t make any difference on Java 8 but will on Java 11 (if you run the new test you added under Java 11, you should see that it fails currently).”

I should have said: “It won’t make any difference on Java 8 but will on Java 11 (if you run the new test you added _with the modification I suggested_ under Java 11, you should see that it fails currently).”

pom.xml
```diff
@@ -139,6 +139,11 @@
             <artifactId>utils</artifactId>
             <version>${pay-java-commons.version}</version>
         </dependency>
+        <dependency>
+            <groupId>uk.gov.pay</groupId>
+            <artifactId>model</artifactId>
+            <version>${pay-java-commons.version}</version>
```
alexbishop1: May as well update the version to the latest 1.0.20190121130619

alexbishop1:
Bah! I got the test thing wrong! The new test will behave the same under Java 8 and Java 11.

However, if the input time came from the system clock, it would behave differently (because Java 8 always returns the current time with millisecond precision while Java 11 uses the best precision available — see [JDK-8068730](https://bugs.openjdk.java.net/browse/JDK-8068730)).

Sorry for the confusion!

PR #388, opened 2019-03-26
(DO NOT MERGE) PP 5020 migrate to java 11

danworth:
## WHAT YOU DID
Updated the base image to use AdoptOpenJDK11 which is based on Alpine.
Update the `pom.xml` to replace `source` and `target` with new `release` flag and set it to 11.
Set `JAVA_HOME` to reference the 11 jdk on CI.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM govukpay/openjdk:alpine-3.9-jre-base-8.201.08
+FROM adoptopenjdk/openjdk11:jdk-11.0.2.9-alpine
 
 RUN apk --no-cache upgrade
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM govukpay/openjdk:alpine-3.9-jre-base-8.201.08
+FROM adoptopenjdk/openjdk11:jdk-11.0.2.9-alpine
```
alexbishop1: We should ascertain what the differences are between the `govukpay` OpenJDK Alpine image and the standard OpenJDK Alpine image e.g. does it have extra packages installed? We probably want to replicate any differences with the AdoptOpenJDK Alpine image.

danworth: 👍 yes, sorry I should have added a note that I raised this PR to see if it would work and investigate any issues. They'll be more rigour around switching the base image before I actually do it.

alexbishop1: In that case, can we try using the slim images?

danworth:
I'm closing this PR whilst we build a proper Alpine OpenJDK base image reflective of our existing one (CI fails because of this) and will raise again as a "draft" PR as suggested by Steven when the time comes.

PR #392, opened 2019-04-02
TBD Convert create gateway account end-point to use validated type

danworth:
- Add `CreateGatewayAccountRequest` with validation mirroring `CreateGatewayAccountValidator` and types reflective of `GatewayAccountParser`.
- Change `GatewayAccountService.create` to use `CreateGatewayAccountRequest` type.
- Modify overloaded `GatewayAccount` constructor to permit passing of `accessToken`
and `organisation`.
- Modify GatewayAccountResource POST end-point to accept a valid `CreateGatewayAccountRequest`.
- Add JsonMappingExceptionMapper with priority over Jackson default to return
`application/json` rather than `text/plain`.
- Annotate `fromString()` method in enums with `@JsonCreator` so that Jackson uses
    these methods when deserialising.
- Remove clases and methods no longer required, e.g. `GatewayAccountParser` and associated tests.
- Add tests for `CreateGatewayAccountRequest` validation and to ensure sensible
error response are sent by the end-point.

## WHAT YOU DID
We already use validated request types as arguments to our end-points in Connector. This provides an implementation within DDConnector for consideration. The benefits are:
- Easier to understand what a valid request looks like
- Easier to add parameters and further validation
- Removal of boiler plate code to do our own parsing and validation.
- More aligned to Java EE and Dropwizard standards making it easier for new joiners to understand and for our projects to follow industry best-practice. Validation is covered by JSR380.

The negatives:
- Although the annotations are well named and their intentions are clear, people unfamiliar with how annotations work may not immediately understand how they take affect.
- The JSON marshalling is done by Jackson and understanding how errors propagate is not as simple when manually parsing the request and throwing the exceptions in our own source code.

On reflection I believe both negative points can be overcome by spending some time exploring how annotations and Jackson works (it is all source code and be explored/understood, there is no compiler magic), and the positives of code that is easier to comprehend, strongly typed, aligned with best-practice and simpler to modify is worth it.

## Reviewer Notes
As discussed above, we follow this pattern in Connector and I believe it is a worthwhile pattern to propagate to other projects. I'm sure there are technical details that can be improved upon and I welcome those comments, but also please consider the general pattern of using Java Bean Validation and typed request parameters within in our end-points rather than `Map<String, String>`.

_What about `@JsonSnakeCase`?_
Some may wonder why we can't use `@JsonSnakeCase` on `CreateGatewayAccountRequest`. I have purposefully made this class immutable, as such each parameter in the public constructor must be annotated with `@JsonProperty`. This is because by default the argument names are not kept post compilation (can be changed post java 8 using `-parameters` flag) and therefor not available to Jackson via reflection at run-time. The annotations on each parameter are however kept and so provides the means for Jackson to know in which order to pass the values.

what is the `@Priority(1)`?
Jackson comes with its own `ExceptionMapper<JsonMappingException>` which returns the response as `text/plain` for any `ProcessingExceptions`. This occurs for example if a request has a String to be converted to an enum which is not valid. `@Priority(1)` gives our implementation priority so that the response is returned as `application/json` following our usual standard.


src/test/java/uk/gov/pay/directdebit/gatewayaccounts/resources/GatewayAccountResourceIT.java
```diff
@@ -284,18 +323,37 @@ public void shouldReturnBadRequestIfValidationFails() throws JsonProcessingExcep
                 .post(GATEWAY_ACCOUNTS_API_PATH)
                 .then()
                 .contentType(JSON)
-                .statusCode(Response.Status.BAD_REQUEST.getStatusCode())
-                .body("message", is("Field(s) missing: [service_name]"));
+                .statusCode(422)
```
mrlumbu: not sure why we return a 422 here and then 400 below, in near identical situation

src/main/java/uk/gov/pay/directdebit/common/exception/JsonMappingExceptionMapper.java
```diff
@@ -0,0 +1,25 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Priority;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+@Priority(1)
```
mrlumbu: this is good to know, and something we should probably stick in all our apps - I have had difficulties debugging stuff on a number of occasions because the default `JsonProcessingException` handling appears totally silent

src/main/java/uk/gov/pay/directdebit/gatewayaccounts/api/CreateGatewayAccountRequest.java
```diff
@@ -0,0 +1,75 @@
...
+    private final PaymentProvider paymentProvider;
+
+    @NotEmpty
+    @Size(min = 1, max = 50, message = "service_name must be between {min} and {max} characters.")
+    private final String serviceName;
+
+    @NotNull
+    private final Type type;
+
+    private final String description;
```
mrlumbu: this looks great, much cleaner. It would be good (maybe in separate PR) to tighten up validation here - all these params must have a max length, and it would be good to catch them here rather than later

src/test/java/uk/gov/pay/directdebit/gatewayaccounts/resources/GatewayAccountResourceIT.java
danworth: Its a good question. As some context the process as I understand it is that the JSON is first deserialised into an instance of `CreateGatewayAccountRequest` and then validation is performed to ensure all constraints are met. 

The `422` is out-of-the-box behaviour as part of the validation step. The request was successfully deserialised into an instance of `CreateGatewayAccountRequest` and then each value is validated, any constraint validations then returned with `422` since `indicates that the server understands the content type of the request entity, and the syntax of the request entity is correct, but it was unable to process the contained instructions`.

The case where an invalid enum value is provided causes an exception that prevents deserialising into an instance of `CreateGatewayAccountRequest` and is thrown as part of the deserialisation and before the validation occurs. Again the out-of-the-box code used by Jackson for this scenario `400` which signals `that the server cannot or will not process the request due to something that is perceived to be a client error`.

The problem is that in both instances from a client's perspective the problem is an invalid value, not an invalid request format in the scenario below, however they receive slightly different error codes. It might be difficult to resolve this since being unable to deserialise into the target class results in a `JsonProcessingException` being thrown within the Jackson library, which we catch with the `JsonProcessingExceptionMapper` included herein. However, it wouldn't necessarily be appropriate to make that mapper return `422` because it would also catch the scenario when someone sends some nonsense for which `400` would be appropriate.  This could probably be resolved somehow...but I didn't spend too much time on it because its an internal api and they both signify and error with a meaningful message.

src/main/java/uk/gov/pay/directdebit/gatewayaccounts/api/CreateGatewayAccountRequest.java
danworth: yes good idea. I've added for `description` and `analysticsId` which are Strings stored into `VARCHAR(255)`. And in this new format, it was super easy to do :)

mrlumbu: 🎉

PR #398, opened 2019-04-08
Remove dependency-check plugin from POM

alexbishop1:
We only ever run this manually, which we can do without it being in the POM:

```
$ mvn org.owasp:dependency-check-maven:check
```


alexbishop1:
We don’t run it routinely. But it’s handy if we want to check for vulnerabilities (perhaps we heard about one and wanted to check if we were affected).

PR #420, opened 2019-05-01
PP-5142 Include error_identifier in error responses

stephencdaly:
- Make it so that all error responses include a "message" field, the value of which is an array of strings and and "error_identifier" field.
- For now, the "error_identifier" is always GENERIC.

- The ErrorIdentifier enum has been created here for now, but will be moved to pay-java-commons.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/common/model/ErrorResponse.java
+++ b/src/main/java/uk/gov/pay/directdebit/common/model/ErrorResponse.java
@@ -1,0 +1,35 @@
+package uk.gov.pay.directdebit.common.model;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+import static com.fasterxml.jackson.annotation.JsonInclude.Include;
+
+@JsonInclude(Include.NON_NULL)
+public class ErrorResponse {
+    
+    @JsonProperty("error_identifier")
+    private final ErrorIdentifier identifier;
+    
+    @JsonProperty("message")
+    private final List<String> messages;
+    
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/pay/directdebit/common/model/ErrorResponse.java
```diff
@@ -0,0 +1,47 @@
...
+@JsonInclude(Include.NON_NULL)
+public class ErrorResponse {
+    
+    @JsonProperty("error_identifier")
+    private final ErrorIdentifier identifier;
+    
+    @JsonProperty("message")
+    private final List<String> messages;
+    
+    @JsonProperty("reason")
```
alexbishop1: Maybe a `@Nullable` annotation (or `@NotNull` on the other fields) would be useful here

src/main/java/uk/gov/pay/directdebit/common/model/ErrorResponse.java
```diff
@@ -0,0 +1,47 @@
...
+    private final String reason;
```
alexbishop1: So we can have multiple messages but only one reason?

stephencdaly: Hmm this was copied over from connector actually, I think we can dispense with the reason for Direct Debit Connector as it's not used. I wasn't planning on making this model common between micro-services, but should we?

src/main/java/uk/gov/pay/directdebit/common/model/ErrorResponse.java
stephencdaly: I'll look at potentially making this optional if we share the model with connector public API via pay-java-commons. For now I'll just remove the reason field in case we decide against sharing it in the end.

PR #442, opened 2019-05-17
PP-5232 Remove reliance on exception to control logic

danworth:
- Make `DirectDebitStatesGraph` return Optional<State> rather than throwing an
exception if no next state can be found. This exception was being relied upon
to control logic with confusing methods such as `void canUpdateStateFor` which
appears to be asking a question but with a return type of `void` was relying upon
the exception being thrown deeper down.
- Move the throwing of exception higher in the call stack and in the context of
the check that was being performed. e.g canUpdateState() ? do something : throw exception.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/services/OneOffMandateService.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/services/OneOffMandateService.java
@@ -6,6 +6,7 @@ import uk.gov.pay.directdebit.mandate.model.Mandate;
 import uk.gov.pay.directdebit.mandate.model.OneOffConfirmationDetails;
 import uk.gov.pay.directdebit.payers.model.BankAccountDetails;
 import uk.gov.pay.directdebit.payments.api.CreatePaymentRequest;
+import uk.gov.pay.directdebit.payments.exception.InvalidStateTransitionException;
 import uk.gov.pay.directdebit.payments.model.PaymentProviderFactory;
 import uk.gov.pay.directdebit.payments.model.Transaction;
 import uk.gov.pay.directdebit.payments.services.TransactionService;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/directdebit/mandate/services/OneOffMandateService.java
```diff
@@ -39,23 +40,25 @@ public Transaction create(GatewayAccount gatewayAccount, CreatePaymentRequest cr
...
-                                confirmMandateRequest.getAccountNumber(),
-                                confirmMandateRequest.getSortCode()),
-                        transaction
-                );
-
-        mandateStateUpdateService.confirmedOneOffDirectDebitDetailsFor(oneOffConfirmationDetails.getMandate());
-        transactionService.oneOffPaymentSubmittedToProviderFor(
-                transaction,
-                oneOffConfirmationDetails.getChargeDate());
+        if (mandateStateUpdateService.canUpdateStateFor(mandate, DIRECT_DEBIT_DETAILS_CONFIRMED)) {
```
alexbishop1: I wonder if we could do away with `canUpdateStateFor(…)` altogether and have something like:

```java
mandateStateUpdateService.getNextStateForEvent(mandate.getState(), DIRECT_DEBIT_DETAILS_CONFIRMED).ifPresentOrElse(nextState -> {
    //…
}, () -> throw new InvalidStateTransitionException(…));
```

danworth: Like the idea but it doesn't appear possible to `throw` from a `Runnable` which is the second argument to `ifPresentOrElse`. Really need an `ifPresentOrElseThrow`, both exist individually but not together (`ifPresent` itself just returns `void`).

PR #447, opened 2019-05-17
PP-5231 Introduce ServiceMandateReference

alexbishop1:
`ServiceMandateReference` represents the reference assigned to a mandate by a government service (set when the service sends the request to us to create the mandate). It wraps a string.

This isn’t quite as nice as it could be because `CreateMandateRequest` (create an on-demand mandate) and `CreatePaymentRequest` (either create a one-off mandate and payment to be collected from it, or collect a payment from an on-demand mandate) both implement `CreateRequest` and `CreatePaymentRequest` also implements `CollectRequest`, which both define the method `getReference()` to return a `String` object (it represents either a service’s reference for a mandate or a service’s reference for a payment or both at the same time).

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/api/CreateMandateRequest.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/api/CreateMandateRequest.java
@@ -1,8 +1,10 @@
 package uk.gov.pay.directdebit.mandate.api;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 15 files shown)


alexbishop1:
This is failing because creating a mandate without a service reference is apparently allowable and the code that converts from strings into the request to `ServiceMandateReference` blows up if there’s not a reference to convert. Might have to rethink this.

alexbishop1:
Closing for now due to the aforementioned complications.

PR #449, opened 2019-05-19
PP-5231 Introduce MandateBankStatementReference

alexbishop1:
`MandateBankStatementReference` represents the reference for a mandate that may be shown on the paying user’s bank statement. It wraps a string.

Having this type seems particularly valuable because it tends to be called either `mandateReference` or `goCardlessReference` and it’s not particularly obvious these represent the same thing when they’re both plain strings. As a result, introducing this was satisfying and changing the types of variables from strings to this new class until everything compiled was particularly satisfying. Like when you’re cutting wrapping paper and the scissors start to glide and you don’t have to make the snipping action any more.

A few classes that represent JSON responses assumed that fields that now have the type `MandateBankStatementReference` could be null (in `equals(…)` methods etc.): as far as I can tell, this isn’t the case so the they’ve been updated to assume the fields will never be null.


src/main/java/uk/gov/pay/directdebit/mandate/model/MandateBankStatementReference.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Objects;
+
+/**
+ * Reference for a mandate, which may be shown on the paying user’s bank
+ * statement (possibly truncated) when a payment is taken
+ * <br>
+ * Since the BACS rules for valid mandate references are complex, we
+ * generally let GoCardless generate them for us
+ * <br>
+ * This object itself does no validation, so instances are not guaranteed to
```
danworth: Strictly speaking besides ensuring the value is not null.

src/main/java/uk/gov/pay/directdebit/mandate/model/MandateBankStatementReferenceArgumentFactory.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.pay.directdebit.mandate.model;
+
+import org.jdbi.v3.core.argument.AbstractArgumentFactory;
+import org.jdbi.v3.core.argument.Argument;
+import org.jdbi.v3.core.config.ConfigRegistry;
+
+import java.sql.Types;
+
+public class MandateBankStatementReferenceArgumentFactory extends AbstractArgumentFactory<MandateBankStatementReference> {
```
danworth: Wondering if there's a way to create an argument factory that would work for all objects that are just wrapping a String since it's seems its always just a null safe `.toString()`; similar to the `ToStringSerializer` so we wouldn't have to keep creating these for each specific type.

alexbishop1: Yeah, I’ve been wondering about that too: we have several of these argument factories now and they’re all nearly identical. Maybe it’s doable if all the types have a common supertype.

PR #452, opened 2019-05-21
PP-5193 Configure proxy for Jersey client

stephencdaly:
Use proxy properties set on the JVM to configure the proxy for a Jersey Client. This fixes sending requests to GoCardless.

Copied how it is configured in our other microservices.

Already tested this on test-12 and it fixes the problem.


oswaldquek:
This follows the code in https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java#L62-L65 so I'm approving this.

PR #454, opened 2019-05-24
Test: Remove Apache connector

rhowe-gds:


rhowe-gds:
Well it passed integration tests. Was an interesting idea.

PR #457, opened 2019-05-24
PP-5234: Remove references to service name

oswaldquek:
I'm not sure whether to just remove the `@NotEmpty` on CreateGatewayAccountRequest's `serviceName` field first. Given direct debit connector isn't in production i thought maybe we could remove the `serviceName` field as well. This relates to https://github.com/alphagov/pay-selfservice/commit/422511d1e0cfcb37568b390902a8db7ed6eddc66.

### DIFF
--- a/src/test/java/uk/gov/pay/directdebit/gatewayaccounts/resources/GatewayAccountResourceIT.java
+++ b/src/test/java/uk/gov/pay/directdebit/gatewayaccounts/resources/GatewayAccountResourceIT.java
@@ -48,7 +48,6 @@ import static uk.gov.pay.directdebit.util.ResponseContainsLinkMatcher.containsLi
 public class GatewayAccountResourceIT {
 
     private static final PaymentProvider PAYMENT_PROVIDER = PaymentProvider.SANDBOX;
-    private static final String SERVICE_NAME = "alex";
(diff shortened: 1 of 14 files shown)


src/test/java/uk/gov/pay/directdebit/gatewayaccounts/resources/GatewayAccountResourceIT.java
```diff
@@ -305,37 +291,17 @@ public void shouldCreateAGatewayAccountWithMinimumDetails() throws JsonProcessin
...
-                .build());
-
-        givenSetup()
-                .body(postBody)
-                .post(GATEWAY_ACCOUNTS_API_PATH)
-                .then()
-                .contentType(JSON)
-                .statusCode(422)
-                .body("errors[0]", is("serviceName may not be empty"));
-    }
```
alexbishop1: Rather than being deleted, I think this test might want to be replaced by something that checks what happens when validation fails for another field (payment provider or something).

oswaldquek: Good shout. I've amended the test to validate the description length.

PR #462, opened 2019-05-29
PP-5234: Drop service name column

oswaldquek:
Do not merge until https://deploy.deploy.payments.service.gov.uk/job/deploy-ecs-selfservice-staging/parambuild/?RELEASE_NUMBER=981&AWS_PROFILE=staging&DEPLOY_ENV=staging-2 is released


alexbishop1:
Seems to be more than just dropping the column here.

oswaldquek:
@alexbishop1 Fair enough. I've updated the PR. The actual dropping of the column will come in another PR.

PR #466, opened 2019-05-31
PP-5244 Extract all information from GoCardlessEvent

RoryMalcolm:
Now stores all information from GoCardless events in the database with migrations to add columns to achieve this functionality.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/payments/dao/GoCardlessEventDao.java
+++ b/src/main/java/uk/gov/pay/directdebit/payments/dao/GoCardlessEventDao.java
@@ -14,10 +14,52 @@ import uk.gov.pay.directdebit.payments.model.GoCardlessEventIdArgumentFactory;
 @RegisterRowMapper(GoCardlessEventMapper.class)
 public interface GoCardlessEventDao {
 
-    @SqlUpdate("INSERT INTO gocardless_events(internal_event_id, event_id, action, resource_type, json, created_at) VALUES (:eventId, :goCardlessEventId, :action, :resourceType, CAST(:json as jsonb), :createdAt)")
+    @SqlUpdate("INSERT INTO gocardless_events(" +
+            " internal_event_id," +
+            " event_id, action," +
+            " resource_type," +
+            " json," +
+            " details_cause," +
+            " details_description," +
+            " details_origin," +
+            " details_reason_code," +
+            " details_scheme," +
+            " links_mandate," +
+            " links_new_customer_bank_account," +
+            " links_new_mandate," +
+            " links_organisation," +
+            " links_parent_event," +
+            " links_payment," +
+            " links_payout," +
+            " links_previous_customer_bank_account," +
+            " links_refund," +
+            " links_subscription," +
+            " created_at)" +
+            " VALUES (" +
+            " :eventId," +
+            " :goCardlessEventId," +
+            " :action," +
+            " :resourceType," +
+            " CAST(:json as jsonb)," +
+            " :detailsCause," +
+            " :detailsDescription," +
+            " :detailsOrigin," +
+            " :detailsReasonCode," +
+            " :detailsScheme," +
+            " :linksMandate," +
+            " :linksNewCustomerBankAccount," +
+            " :linksNewMandate," +
+            " :linksOrganisation," +
+            " :linksParentEvent," +
+            " :linksPayment," +
+            " :linksPayout," +
+            " :linksPreviousCustomerBankAccount," +
+            " :linksRefund," +
+            " :linksSubscription," +
+            " :createdAt)")
     @GetGeneratedKeys
     Long insert(@BindBean GoCardlessEvent goCardlessEvent);
-    
+
     @SqlUpdate("UPDATE gocardless_events t SET internal_event_id = :eventId WHERE t.id = :id")
     int updateEventId(@Bind("id") Long id, @Bind("eventId") Long eventId);
 }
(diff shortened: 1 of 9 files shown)


src/main/resources/migrations/00045_alter_table_gocardless_events.sql
```diff
@@ -0,0 +1,56 @@
+
```
danworth: The file name should probably begin `00047` as we use a sequential prefix, looks like the next one is 47 maybe someone snuck in new files ahead of you :)

src/main/resources/migrations/00045_alter_table_gocardless_events.sql
```diff
@@ -0,0 +1,56 @@
...
+--changeset uk.gov.pay:alter_table_gocardless_events
```
danworth: We usually have db migration scripts in their own PR, which will also help reduce the size of this one. By default the migration scripts do not run when deploying to staging and production and its necessary to manually invoke them. Would you mind splitting this out into it's own PR please?

src/main/java/uk/gov/pay/directdebit/payments/dao/mapper/GoCardlessEventMapper.java
```diff
@@ -19,6 +19,21 @@
     private static final String RESOURCE_TYPE_COLUMN = "resource_type";
     private static final String JSON_COLUMN = "json";
     private static final String CREATED_AT_COLUMN = "created_at";
+    private static final String DETAILS_CAUSE_COLUMN = "details_cause";
```
danworth: Looks like all of these constants are only ever used within this file and in a single location so I would be inclined to in-line them where they are being used for easier readability.

src/main/java/uk/gov/pay/directdebit/payments/model/GoCardlessEvent.java
```diff
@@ -138,4 +219,139 @@ public GoCardlessOrganisationId getOrganisationIdentifier() {
     public void setOrganisationIdentifier(GoCardlessOrganisationId organisationIdentifier) {
         this.organisationIdentifier = organisationIdentifier;
     }
+
+    public String getDetailsCause() {
```
danworth: Would be worth considering the use of `Optional` to wrap any optional attributes in the return type of the getters, hopefully the goCardless documentation might tell you which fields are optional. That way it'll be clear to anyone using a `GoCardlessEvent` what is optional.

src/main/java/uk/gov/pay/directdebit/payments/model/GoCardlessEvent.java
```diff
@@ -12,9 +12,24 @@
     //todo action should be typed (see https://developer.gocardless.com/api-reference/#events-payment-actions and the equivalent for other resource_types
     private String action;
     private GoCardlessResourceType resourceType;
+    private String resourceId;
```
danworth: I think the `GoCardlessEvent` can be considered immutable, so all fields can be `final` and no need to provide setters.

src/test/java/uk/gov/pay/directdebit/payments/fixtures/GoCardlessEventFixture.java
```diff
@@ -109,6 +125,141 @@ public GoCardlessEventFixture withOrganisationIdentifier(GoCardlessOrganisationI
         return this;
     }
 
+    public String getDetailsCause() {
+        return detailsCause;
+    }
+
+    public GoCardlessEventFixture setDetailsCause(String detailsCause) {
```
danworth: For the setting methods on fixtures and builders we usually use the prefix `with` rather than `set` so it reads nicer:
```
aGoCardlessEventFixture
.withDetailsCause("somecause")
.withDetailsDescription("some description);
```

src/main/java/uk/gov/pay/directdebit/payments/dao/GoCardlessEventDao.java
```diff
@@ -14,10 +14,43 @@
 @RegisterRowMapper(GoCardlessEventMapper.class)
 public interface GoCardlessEventDao {
 
-    @SqlUpdate("INSERT INTO gocardless_events(internal_event_id, event_id, action, resource_type, json, created_at) VALUES (:eventId, :goCardlessEventId, :action, :resourceType, CAST(:json as jsonb), :createdAt)")
+    @SqlUpdate("INSERT INTO gocardless_events(internal_event_id, event_id, action, resource_type, json," +
```
danworth: There is a test in `GoCardlessDirectDebitEventDaoIT.shouldInsertAnEvent` which should be extended to assert that the new values are being inserted and retrieved correctly.

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/api/GoCardlessWebhookParser.java
```diff
@@ -32,16 +32,63 @@ public GoCardlessWebhookParser(ObjectMapper objectMapper) {
...
+                        = detailsNode.has("payment") ? linksNode.get("payment").asText() : null;
+                String linksPayout 
+                        = detailsNode.has("payout") ? linksNode.get("payout").asText() : null;
+                String linksPreviousCustomerBankAccount = detailsNode.has("previous_customer_bank_account") 
+                        ? linksNode.get("previous_customer_bank_account").asText() : null;
+                String linksRefund 
+                        = detailsNode.has("refund") ? linksNode.get("refund").asText() : null;
+                String linksSubscription = detailsNode.has("subscription") 
+                        ? linksNode.get("subscription").asText() : null;
+                        String resourceType = eventNode.get("resource_type").asText();
```
danworth: Since there are three overloaded constructors for `GoCardlessEvent` each taking about 20 parameters and many being optional perhaps now is a good time to introduce a builder for `GoCardlessEvent`?

src/main/java/uk/gov/pay/directdebit/payments/model/GoCardlessEvent.java
```diff
@@ -12,9 +12,24 @@
...
     private String json;
+    private String detailsCause;
```
danworth: Has there been any discussion whether to create types for the new properties rather than using just using `String`? @alexbishop1 has done some great work to remove using `String` where it makes sense and improves readability and the best time to introduce it would be when we first add the new parameters.

PR #471, opened 2019-06-03
PP-5243: Delete MandateServiceFactory

oswaldquek:
It's quite a pointless class.


src/main/java/uk/gov/pay/directdebit/mandate/resources/MandateResource.java
```diff
@@ -34,13 +35,14 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(MandateResource.class);
     private final MandateService mandateService;
     private final CreateMandateRequestValidator createMandateRequestValidator = new CreateMandateRequestValidator();
-    private final MandateServiceFactory mandateServiceFactory;
+    private final OnDemandMandateService onDemandMandateService;
```
danworth: Are we at the point where this can be renamed without the `OnDemand` prefix?

oswaldquek: good spot. yes we are. i'll delete it.

oswaldquek: oh actually we aren't. it's used by MandateResource.

PR #472, opened 2019-06-03
Pp 5243 delete moar stuff

oswaldquek:


src/main/java/uk/gov/pay/directdebit/payments/services/DirectDebitEventService.java
```diff
@@ -80,9 +80,9 @@ public DirectDebitEvent registerMandateCancelledEventFor(Mandate mandate) {
         return insertEventFor(mandate, directDebitEvent);
     }
 
-    public DirectDebitEvent registerMandateExpiredEventFor(Mandate mandate) {
+    public void registerMandateExpiredEventFor(Mandate mandate) {
```
danworth: Nice cleaning up of unused return value. Looking at the other methods in `DirectDebitEventService` I think most (perhaps all) can be changed to `void`; seems to be some minor use but in tests or with a calling method returning the returned `DirectDebitEvent` but then that not being used anyway. Happy to approve as-is but might be nice to clean them all up if appropriate and if you have time.

PR #479, opened 2019-06-04
PP-5244 Sandbox Events Storage

RoryMalcolm:
- Create SandboxEvent class with mapper and dao
- Update Sandbox webhook resource to work with SandboxEvent
- Add test for SandboxEventDaoIT.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/webhook/sandbox/resources/WebhookSandboxResource.java
+++ b/src/main/java/uk/gov/pay/directdebit/webhook/sandbox/resources/WebhookSandboxResource.java
@@ -1,16 +1,18 @@
 package uk.gov.pay.directdebit.webhook.sandbox.resources;
 
-import java.util.List;
+import com.codahale.metrics.annotation.Timed;
+import uk.gov.pay.directdebit.gatewayaccounts.model.PaymentProvider;
+import uk.gov.pay.directdebit.payments.model.PaymentState;
+import uk.gov.pay.directdebit.events.model.SandboxEvent;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/directdebit/webhook/sandbox/resources/WebhookSandboxResource.java
```diff
@@ -1,36 +1,60 @@
...
 
     @Inject
-    public WebhookSandboxResource(TransactionService transactionService) {
+    public WebhookSandboxResource(TransactionService transactionService, SandboxEventService sandboxEventService) {
         this.transactionService = transactionService;
+        this.sandboxEventService = sandboxEventService;
+    }
+    
+    enum SandboxEventAction {
+        SANDBOX_EVENT_ACTION
```
alexbishop1: I think this (and the cause) should probably be `PAID_OUT` or something to reflect what’s already there

danworth: Good idea I've updated.

danworth:
> We might want to also think about having an integration test for the `WebhookSandboxResource`.

Shall we wait until we've expanded on what we want the Sandbox webhook to do before adding a ResourceIT (we could add as part of the events work we're planning once have a better idea)? I'd like to get this PR merged before end of sprint so we've got the Sandbox events ready for the next stories.

alexbishop1:
Sure, we can wait until we have more significant functionality before adding an integration test.

PR #485, opened 2019-06-07
PP-5266: Add flag to indicate whether account is linked

oswaldquek:
At this moment in time when we say linked, it's whether it's linked
specifically to a GoCardless account.

A bit tangential to the purpose of the PR but some test code cleanup in this PR
too.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/gatewayaccounts/api/GatewayAccountResponse.java
+++ b/src/main/java/uk/gov/pay/directdebit/gatewayaccounts/api/GatewayAccountResponse.java
@@ -33,16 +33,25 @@ public class GatewayAccountResponse {
     private String type;
     @JsonProperty("analytics_id")
     private String analyticsId;
+    @JsonProperty("is_connected")
+    private final boolean isConnected;
     @JsonProperty("links")
     private List<Map<String, Object>> links = new ArrayList<>();
 
-    public GatewayAccountResponse(Long gatewayAccountId, String gatewayAccountExternalId, String paymentProvider, String description, String type, String analyticsId) {
+    public GatewayAccountResponse(Long gatewayAccountId, 
+                                  String gatewayAccountExternalId, 
+                                  String paymentProvider, 
+                                  String description, 
+                                  String type, 
+                                  String analyticsId, 
+                                  boolean isConnected) {
         this.gatewayAccountId = gatewayAccountId;
         this.gatewayAccountExternalId = gatewayAccountExternalId;
         this.paymentProvider = paymentProvider;
         this.description = description;
         this.type = type;
         this.analyticsId = analyticsId;
+        this.isConnected = isConnected;
     }
 
     public static GatewayAccountResponse from(GatewayAccount gatewayAccount) {
@@ -52,8 +61,8 @@ public class GatewayAccountResponse {
                 gatewayAccount.getPaymentProvider().toString(),
                 gatewayAccount.getDescription(),
                 gatewayAccount.getType().toString(),
-                gatewayAccount.getAnalyticsId()
-        );
+                gatewayAccount.getAnalyticsId(),
+                gatewayAccount.getAccessToken().isPresent());
     }
 
     public GatewayAccountResponse withSelfLink(UriInfo uriInfo) {
--- a/src/test/java/uk/gov/pay/directdebit/gatewayaccounts/resources/GatewayAccountResourceIT.java
+++ b/src/test/java/uk/gov/pay/directdebit/gatewayaccounts/resources/GatewayAccountResourceIT.java
@@ -3,7 +3,6 @@ package uk.gov.pay.directdebit.gatewayaccounts.resources;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.collect.ImmutableMap;
 import io.restassured.response.ValidatableResponse;
 import io.restassured.specification.RequestSpecification;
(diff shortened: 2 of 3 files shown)


src/main/java/uk/gov/pay/directdebit/gatewayaccounts/api/GatewayAccountResponse.java
```diff
@@ -33,16 +35,25 @@
     private String type;
     @JsonProperty("analytics_id")
     private String analyticsId;
+    @JsonProperty("is_linked")
+    private final boolean isLinked;
```
alexbishop1: I wonder if `is_connected` might be better for this? GoCardless refer to them as connected accounts (and I think Stripe do something similar and consistency is always nice).

src/main/java/uk/gov/pay/directdebit/gatewayaccounts/api/GatewayAccountResponse.java
```diff
@@ -52,8 +63,8 @@ public static GatewayAccountResponse from(GatewayAccount gatewayAccount) {
                 gatewayAccount.getPaymentProvider().toString(),
                 gatewayAccount.getDescription(),
                 gatewayAccount.getType().toString(),
-                gatewayAccount.getAnalyticsId()
-        );
+                gatewayAccount.getAnalyticsId(),
+                gatewayAccount.getAccessToken().map(token -> TRUE).orElse(FALSE));
```
alexbishop1: We can just do `gatewayAccount.getAccessToken().isPresent()`.

src/test/java/uk/gov/pay/directdebit/gatewayaccounts/resources/GatewayAccountResourceIT.java
```diff
@@ -89,6 +92,32 @@ public void shouldReturnAGatewayAccountIfItExists() {
                 .body(TYPE_KEY, is(TYPE.toString()))
                 .body(EXTERNAL_ID_KEY, is(EXTERNAL_ID))
                 .body(DESCRIPTION_KEY, is(DESCRIPTION))
+                .body(IS_LINKED, is(true))
+                .body(ANALYTICS_ID_KEY, is(ANALYTICS_ID));
+    }
+    
+    @Test
+    public void isLinkedShouldBeFalseIfGatewayAccountHasNoAccessToken() {
```
alexbishop1: Should there also be a test that checks it’s linked if there is an access token?

src/test/java/uk/gov/pay/directdebit/gatewayaccounts/services/GatewayAccountServiceTest.java
```diff
@@ -57,13 +56,13 @@
     private MandateFixture mandateFixture = MandateFixture.aMandateFixture()
             .withGatewayAccountFixture(gatewayAccountFixture);
     @Mock
-    private GatewayAccountDao mockedGatewayAccountDao;
+    private GatewayAccountDao gatewayAccountDao;
```
alexbishop1: It’s a matter of taste but I actually prefer having a prefix on mock objects (though I prefer “mock” to “mocked”) because it makes it really clear we’re dealing with a mock object

src/main/java/uk/gov/pay/directdebit/gatewayaccounts/api/GatewayAccountResponse.java
oswaldquek: lol. of course. doh.

src/test/java/uk/gov/pay/directdebit/gatewayaccounts/resources/GatewayAccountResourceIT.java
oswaldquek: i added a test a few lines above (see line 95)

src/test/java/uk/gov/pay/directdebit/gatewayaccounts/services/GatewayAccountServiceTest.java
oswaldquek: I guess I was finding it really obvious it was a mock as mentions of gatewayAccountDao are usually preceded by "when"s, for e.g. `when(gatewayAccountDao.findByExternalId(accountId)).thenReturn`...

src/main/java/uk/gov/pay/directdebit/gatewayaccounts/api/GatewayAccountResponse.java
oswaldquek: Agreed. I've renamed this.

src/test/java/uk/gov/pay/directdebit/gatewayaccounts/services/GatewayAccountServiceTest.java
alexbishop1: I appreciate that. Though I do like how in a line like `when(mockGatewayAccountDao.findByExternalid(accountId).thenReturn(mockGatewayAccount)` the prefix makes it clear which variables are mocks and which aren’t. I think a prefix is particularly useful when _writing_ (rather than reading) code because I sometimes lose track of what’s a mock and what isn’t.

I think we should probably have a team standard for this but we don’t have one right now, so I’m happy to go with what you prefer in this PR.

src/test/java/uk/gov/pay/directdebit/gatewayaccounts/resources/GatewayAccountResourceIT.java
alexbishop1: Ah, missed that, sorry. It wasn’t too noticeable from the diff.

PR #486, opened 2019-06-10
PP-5278 Don't allow mandate creation if GoCardless account isn't linked

RoryMalcolm:
- Introduces new exceptions to be thrown if a mandate is created by a user who does not have a linked GoCardless account

- Removes access token from config files to ensure that access token is never used as a fallback

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/DirectDebitConnectorApp.java
+++ b/src/main/java/uk/gov/pay/directdebit/DirectDebitConnectorApp.java
@@ -25,8 +25,10 @@ import uk.gov.pay.directdebit.common.exception.BadRequestExceptionMapper;
 import uk.gov.pay.directdebit.common.exception.ConflictExceptionMapper;
 import uk.gov.pay.directdebit.common.exception.InternalServerErrorExceptionMapper;
 import uk.gov.pay.directdebit.common.exception.JsonMappingExceptionMapper;
+import uk.gov.pay.directdebit.common.exception.NoAccessTokenExceptionMapper;
 import uk.gov.pay.directdebit.common.exception.NotFoundExceptionMapper;
 import uk.gov.pay.directdebit.common.exception.PreconditionFailedExceptionMapper;
+import uk.gov.pay.directdebit.common.exception.UnlinkedGCMerchantAccountExceptionMapper;
 import uk.gov.pay.directdebit.common.proxy.CustomInetSocketAddressProxySelector;
 import uk.gov.pay.directdebit.events.resources.DirectDebitEventsResource;
 import uk.gov.pay.directdebit.gatewayaccounts.GatewayAccountParamConverterProvider;
@@ -122,7 +124,8 @@ public class DirectDebitConnectorApp extends Application<DirectDebitConfig> {
         environment.jersey().register(new InternalServerErrorExceptionMapper());
         environment.jersey().register(new PreconditionFailedExceptionMapper());
         environment.jersey().register(new JsonMappingExceptionMapper());
-
+        environment.jersey().register(new NoAccessTokenExceptionMapper());
+        environment.jersey().register(new UnlinkedGCMerchantAccountExceptionMapper());
         initialiseMetrics(configuration, environment);
     }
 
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/pay/directdebit/DirectDebitConnectorApp.java
```diff
@@ -21,12 +21,7 @@
 import uk.gov.pay.directdebit.app.config.GraphiteConfig;
 import uk.gov.pay.directdebit.app.healthcheck.Database;
 import uk.gov.pay.directdebit.app.healthcheck.Ping;
-import uk.gov.pay.directdebit.common.exception.BadRequestExceptionMapper;
-import uk.gov.pay.directdebit.common.exception.ConflictExceptionMapper;
-import uk.gov.pay.directdebit.common.exception.InternalServerErrorExceptionMapper;
-import uk.gov.pay.directdebit.common.exception.JsonMappingExceptionMapper;
-import uk.gov.pay.directdebit.common.exception.NotFoundExceptionMapper;
-import uk.gov.pay.directdebit.common.exception.PreconditionFailedExceptionMapper;
+import uk.gov.pay.directdebit.common.exception.*;
```
stephencdaly: We don't like wildcard imports, you probably need to change your intelliJ settings then re-run optimise imports.

https://gds-way.cloudapps.digital/manuals/programming-languages/java.html#imports

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateService.java
```diff
@@ -78,35 +80,40 @@ public MandateService(DirectDebitConfig directDebitConfig,
...
-                    LOGGER.error("Gateway account with id {} not found", accountExternalId);
-                    return new GatewayAccountNotFoundException(accountExternalId);
-                });
+        Optional<GatewayAccount> potentialGatewayAccount = gatewayAccountDao.findByExternalId(accountExternalId);
+
+        if (potentialGatewayAccount.isPresent()) {
+            if (potentialGatewayAccount.get().getAccessToken().isEmpty()) {
+                LOGGER.error("Gateway account with id {} does not have access token", accountExternalId);
+                throw new UnlinkedGCMerchantAccountException(accountExternalId);
+            }
```
stephencdaly: I think this can be inside the `.map`.
Generally we try to avoid using `.isPresent` as there's usually a more functional option, like using `.map` and `.orElseThrow` as we do here.

RoryMalcolm:
> Does what's required, just some comments about style.

Done!

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateService.java
```diff
@@ -77,36 +80,33 @@ public MandateService(DirectDebitConfig directDebitConfig,
...
-                });
+    Mandate createMandate(CreateMandateRequest createRequest, String accountExternalId) {
+        return gatewayAccountDao.findByExternalId(accountExternalId).map(gatewayAccount -> {
+            if (gatewayAccount.getAccessToken().isEmpty()) {
+                LOGGER.error("Gateway account with id {} has no access token", accountExternalId);
+                throw new NoAccessTokenException(accountExternalId);
+            }
+            MandateBankStatementReference mandateReference = MandateBankStatementReference.valueOf(
+                    PaymentProvider.SANDBOX.equals(gatewayAccount.getPaymentProvider()) ?
+                            RandomStringUtils.randomAlphanumeric(18) : "gocardless-default");
```
stephencdaly: Looks like there was a TODO to set the mandate reference for gocardless accounts here. Not related to what you were doing but we should check if we need a story for this.

stephencdaly:
Can we add a test that the appropriate exceptions are thrown?

RoryMalcolm:
@stephencdaly Done

src/test/java/uk/gov/pay/directdebit/mandate/services/MandateServiceTest.java
```diff
@@ -221,6 +224,38 @@ public void confirm_shouldNotConfirmOnDemandMandateForInvalidState() {
...
+                .withAccessToken(null)
+                .toEntity();
+        when(gatewayAccountDao.findByExternalId(gatewayAccount.getExternalId())).thenReturn(Optional.of(gatewayAccount));
+        boolean exceptionThrown = false;
+        try {
+            service.createMandate(null, gatewayAccount.getExternalId());
+        } catch (UnlinkedGCMerchantAccountException e) {
+            exceptionThrown = true;
+        }
+        assertThat(exceptionThrown, is(true));
```
stephencdaly: Does the job, but we tend to use 
```
@Rule
public ExpectedException thrown = ExpectedException.none();
```
at the top of the class, then in the test we can do
```
thrown.expect(UnlinkedGcMerchantAccountException.class);
```

RoryMalcolm: Fixed both tests :)

PR #487, opened 2019-06-10
PP-5246 No longer persist GoCardlessMandate

danworth:
No longer persist `GoCardlessMandate` when confirming a mandate and prepare for
deletion of `GoCardlessMandate` class, service and Dao. Includes:
- Abstract Mandate interface from existing class Mandate and rename that class
to `MandateImpl`.
- Create `GoCardlessMandateAdaptor` which implements Mandate interface. This class
adapts the gocardless mandate to our model of Mandate.
- Rename Mandate field `mandateReference` to `mandateBankStatementReference` to help
clarify between the other `serviceReference` field.
- No longer persist the gocardlessMandate after confirmation with goCardless.
- Update the `PaymentProviderMandateId` when updating the mandate after confirmation.


danworth:
Might have a simpler way of doing this, closing this PR whilst I try it out.

PR #488, opened 2019-06-10
PP-5246 No longer persist GoCardlessMandate

danworth:
No longer persist GoCardlessMandate when confirming a mandate and prepare for
    deletion of GoCardlessMandate class, service and Dao. Includes:
    - Rename Mandate field mandateReference to mandateBankStatementReference to help
    clarify between the other serviceReference field.
    - No longer persist the gocardlessMandate after creation with GoCardless.
    - No longer create a GoCardlessMandate within the GoCardlessClientFacade, instead
    return a new PaymentProviderConfirmMandateResponse which includes the GoCardless
    mandate Id and the bank statement reference.

Update:
@alexbishop1 I have introduced a `PaymentProviderConfirmMandateResponse` to avoid mutating a `Mandate` object within the lower level GoCardless/Sandbox services. I think this makes it clearer and leaves the logic to mutate the Mandate and persist changes within the MandateService layer. I eagerly await your opinions...

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/common/clients/GoCardlessClientFacade.java
+++ b/src/main/java/uk/gov/pay/directdebit/common/clients/GoCardlessClientFacade.java
@@ -13,6 +13,7 @@ import uk.gov.pay.directdebit.mandate.model.GoCardlessMandateId;
 import uk.gov.pay.directdebit.mandate.model.GoCardlessPayment;
 import uk.gov.pay.directdebit.mandate.model.Mandate;
 import uk.gov.pay.directdebit.mandate.model.MandateBankStatementReference;
+import uk.gov.pay.directdebit.mandate.model.PaymentProviderMandateIdAndBankReference;
 import uk.gov.pay.directdebit.mandate.model.subtype.MandateExternalId;
 import uk.gov.pay.directdebit.payers.model.AccountNumber;
 import uk.gov.pay.directdebit.payers.model.BankAccountDetails;
@@ -50,12 +51,11 @@ public class GoCardlessClientFacade {
         return customer;
     }
 
-    public GoCardlessMandate createMandate(Mandate mandate, GoCardlessCustomer customer) {
-        com.gocardless.resources.Mandate gcMandate = goCardlessClientWrapper.createMandate(mandate.getExternalId(), customer);
-        return new GoCardlessMandate(
-                mandate.getId(),
-                GoCardlessMandateId.valueOf(gcMandate.getId()),
-                MandateBankStatementReference.valueOf(gcMandate.getReference()));
+    public PaymentProviderMandateIdAndBankReference createMandate(Mandate mandate, GoCardlessCustomer customer) {
+        var createdMandate = goCardlessClientWrapper.createMandate(mandate.getExternalId(), customer);
+        return new PaymentProviderMandateIdAndBankReference(
+                GoCardlessMandateId.valueOf(createdMandate.getId()),
+                MandateBankStatementReference.valueOf(createdMandate.getReference()));
     }
 
     public GoCardlessPayment createPayment(Transaction transaction, GoCardlessMandate mandate) {
--- a/src/main/java/uk/gov/pay/directdebit/mandate/model/Mandate.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/model/Mandate.java
@@ -14,23 +14,23 @@ public class Mandate {
     private MandateState state;
     private final GatewayAccount gatewayAccount;
     private final String returnUrl;
-    private MandateBankStatementReference mandateReference;
+    private MandateBankStatementReference mandateBankStatementReference;
     private final String serviceReference;
     private final ZonedDateTime createdDate;
     private Payer payer;
-    private PaymentProviderMandateId paymentProviderId;
+    private PaymentProviderMandateId paymentProviderMandateId;
 
     private Mandate(MandateBuilder builder) {
         this.id = builder.id;
         this.gatewayAccount = builder.gatewayAccount;
         this.externalId = builder.externalId;
-        this.mandateReference = builder.mandateReference;
+        this.mandateBankStatementReference = builder.mandateBankStatementReference;
         this.serviceReference = builder.serviceReference;
         this.state = builder.state;
         this.returnUrl = builder.returnUrl;
         this.createdDate = builder.createdDate;
         this.payer = builder.payer;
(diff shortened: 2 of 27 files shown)


src/main/java/uk/gov/pay/directdebit/common/clients/GoCardlessClientFacade.java
```diff
@@ -50,12 +50,11 @@ public GoCardlessCustomer createCustomerBankAccount(MandateExternalId mandateExt
         return customer;
     }
 
-    public GoCardlessMandate createMandate(Mandate mandate, GoCardlessCustomer customer) {
+    public Mandate createMandate(Mandate mandate, GoCardlessCustomer customer) {
```
alexbishop1: Maybe this method should be `void` and called something other than `createMandate` so that it’s obvious it’s modifying the passed-in `Mandate` rather than creating something new?

I suppose this method _could_ return an entirely new `Mandate` — I’m not sure what will make it clearer than the `Mandate` needs to be persisted (whether it’s the modified original or a new one).

Part of me is now wondering whether we should have some sort of inheritance relationship whereby there’s a `Mandate` that has a `MandateThatExistsWithAPaymentProvider` (stupid name just to illustrate the point).

src/main/java/uk/gov/pay/directdebit/mandate/model/Mandate.java
```diff
@@ -69,24 +69,24 @@ public void setState(MandateState state) {
         this.state = state;
     }
 
-    public MandateBankStatementReference getMandateReference() {
-        return mandateReference;
+    public MandateBankStatementReference getMandateBankStatementReference() {
```
alexbishop1: Should this maybe return an `Optional`? (Possibly not a change for now)

src/main/java/uk/gov/pay/directdebit/common/clients/GoCardlessClientFacade.java
danworth: Yes its a good point and something that I'm not happy with myself. If the only thing that needed to be persisted on our mandate after the creation of the mandate with GoCardless was their Id, then we could make `createMandate` simply return the `PaymentProviderMandateId` back to the MandateService for it to persist (which I think is more intuitive). I'm not sure why we take the bank statement reference at the same time we create the mandate with go cardless as I suspect they take it from something we give them and which we might know earlier in the process. I'll have a look around and see if I can make it cleaner.

src/main/java/uk/gov/pay/directdebit/mandate/model/Mandate.java
danworth: Yes it should but it was existing and didn't want to play around too much at the same time. I'll either add it to this PR after addressing your comment above or in a future one.

danworth:
> This looks mostly straightforward.
> 
> I can’t see any code to update the existing mandate in the database after it’s been created with GoCardless. Is that supposed to be in this PR or is it planned for a future one?

It happens here
```
private Mandate confirmedDetailsFor(Mandate mandate) {	        
        updateStateFor(mandate, DIRECT_DEBIT_DETAILS_CONFIRMED);
        mandateDao.updateReferenceAndPaymentProviderId(mandate);
        return mandate;
}
```

src/main/java/uk/gov/pay/directdebit/common/clients/GoCardlessClientFacade.java
alexbishop1: The reason the bank statement reference comes from GoCardless is because they generate it for us (the Bacs rules for what’s valid are complicated so we let them worry about getting it right).

danworth: 👍 yep, I got that from their docs

src/main/java/uk/gov/pay/directdebit/common/clients/GoCardlessClientFacade.java
```diff
@@ -50,12 +51,11 @@ public GoCardlessCustomer createCustomerBankAccount(MandateExternalId mandateExt
...
     }
 
-    public GoCardlessMandate createMandate(Mandate mandate, GoCardlessCustomer customer) {
-        com.gocardless.resources.Mandate gcMandate = goCardlessClientWrapper.createMandate(mandate.getExternalId(), customer);
-        return new GoCardlessMandate(
-                mandate.getId(),
-                GoCardlessMandateId.valueOf(gcMandate.getId()),
-                MandateBankStatementReference.valueOf(gcMandate.getReference()));
+    public PaymentProviderConfirmMandateResponse createMandate(Mandate mandate, GoCardlessCustomer customer) {
+        var createdMandate = goCardlessClientWrapper.createMandate(mandate.getExternalId(), customer);
```
alexbishop1: Avoiding having a huge fully-qualified class name in the type declaration by using `var`… sneaky! You have my admiration!

src/test/java/uk/gov/pay/directdebit/payments/services/GoCardlessServiceOnDemandTest.java
```diff
@@ -30,42 +35,56 @@ public void setUp() {
...
-    public void confirm_shouldStoreAGoCardlessCustomerBankAccountMandateAndPaymentWhenReceivingConfirmTransaction() {
+    public void confirm_shouldCreateAGoCardlessCustomerBankAccountMandateAndPaymentWhenReceivingConfirmTransaction() {
         Mandate mandate = mandateFixture.toEntity();
+        var confirmMandateResponse = 
+                new PaymentProviderConfirmMandateResponse(
+                        SandboxMandateId.valueOf("aPaymentProviderId"),
+                        MandateBankStatementReference.valueOf(BANK_ACCOUNT_ID));
 
-        when(mockedGoCardlessClientFacade.createMandate(mandate, goCardlessCustomer)).thenReturn(goCardlessMandate);
+        when(mockedGoCardlessClientFacade.createMandate(mandate, goCardlessCustomer)).thenReturn(confirmMandateResponse);
```
alexbishop1: Can probably just inline the creation of `confirmMandateResponse` here.

danworth:
> PaymentProviderMandateIdAndBankReference

Yes I like that better, I'll change it.

src/test/java/uk/gov/pay/directdebit/payments/services/GoCardlessServiceOnDemandTest.java
danworth: 👍

PR #494, opened 2019-06-11
Pp 5246 Remove GoCardlessMandate and Dao

danworth:
## WHAT YOU DID
Remove GoCardlessMandate and Dao.

- No longer save the GoCardlessMandate after creating it with GoCardless.
- When collecting a payment retrieve the Mandate rather than looking for a GoCardlessMandate, and pass the PaymentProviderMandateId to GoCardlessService which is all it needs from the mandate to create a payment.
- When handling a GoCardlessEvent relating to mandate, look up the mandate via the linksMandate attribute (which is a PaymentProviderMandateId) rather than finding a GoCardlessMandate.
- Delete the GoCardlessMandate, GoCardlessMandateDao and GoCardlessMandateFixture.
- Update tests

## Notes
Unfortunately its on the large side. Previous attempt to do this in stages broke our smoke test since it looks for the GoCardlessMandate when collecting a payment.

@alexbishop1 has already reviewed and approved half of the changes (which were unfortunately reverted to unblock the failing smoke tests). It touches a lot of files but the changes are relatively simple. I'd be happy to help walk through the changes with anyone that can review them.


src/main/java/uk/gov/pay/directdebit/mandate/services/MandateService.java
```diff
@@ -210,15 +211,17 @@ public void cancelMandateCreation(MandateExternalId mandateExternalId) {
...
+                    .confirmMandate(
                             mandate,
                             new BankAccountDetails(
                                     confirmDetailsRequest.getAccountNumber(),
                                     confirmDetailsRequest.getSortCode())
                     );
-            mandateStateUpdateService.confirmedOnDemandDirectDebitDetailsFor(confirmedMandate);
+            mandate.setMandateBankStatementReference(paymentProviderMandateIdAndBankReference.getMandateBankStatementReference());
+            mandate.setPaymentProviderMandateId(paymentProviderMandateIdAndBankReference.getPaymentProviderMandateId());
+            mandateStateUpdateService.confirmedOnDemandDirectDebitDetailsFor(mandate);
```
stephencdaly: "OnDemand" should go from this method name but that can be a separate PR

src/test/java/uk/gov/pay/directdebit/mandate/services/MandateStateUpdateServiceTest.java
```diff
@@ -132,25 +132,13 @@ public void shouldRegisterEventWhenExchangingTokens() {
...
-    }
-
     @Test
     public void shouldUpdateMandateStateAndRegisterEventWhenConfirmingDirectDebitDetails_andSendEmail_IfOnDemand() {
-        Mandate newMandate = service.confirmedOnDemandDirectDebitDetailsFor(onDemandMandate);
+        Mandate confirmedMandate = service.confirmedOnDemandDirectDebitDetailsFor(onDemandMandate);
 
-        assertThat(newMandate, is(onDemandMandate));
+        assertThat(confirmedMandate, is(onDemandMandate));
         verify(mockedUserNotificationService).sendOnDemandMandateCreatedEmailFor(onDemandMandate);
```
stephencdaly: can get rid of all references to "OnDemand" mandates here as there's no longer that distinction. Can be a separate PR

PR #498, opened 2019-06-12
PP-5246 Use SandboxMandateId within SandboxEvent

danworth:
PP-5246 Use SandboxMandateId within SandboxEvent
    
    - Change `mandateId` type from `String` to `SandboxMandateId`
    - Make return type of getPaymentId and getMandateId optional
    
    with @RoryMalcolm

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/events/dao/SandboxEventMapper.java
+++ b/src/main/java/uk/gov/pay/directdebit/events/dao/SandboxEventMapper.java
@@ -1,9 +1,9 @@
 package uk.gov.pay.directdebit.events.dao;
 
+import org.jdbi.v3.core.mapper.RowMapper;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/directdebit/events/dao/SandboxEventMapper.java
```diff
@@ -4,20 +4,27 @@
 import uk.gov.pay.directdebit.events.model.SandboxEvent;
 
 import org.jdbi.v3.core.mapper.RowMapper;
+import uk.gov.pay.directdebit.mandate.model.SandboxMandateId;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.time.ZonedDateTime;
+import java.util.stream.Stream;
```
RoryMalcolm: Unused import

PR #502, opened 2019-06-13
PP-5428 Create types for Sandbox and GoCardless payment ids

danworth:


src/main/java/uk/gov/pay/directdebit/payments/model/GoCardlessPaymentId.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.pay.directdebit.payments.model;
+
+/**
+ * The ID assigned by GoCardless to a Direct Debit payment e.g. "MD123"
```
alexbishop1: This example should be `"PM123"` (GoCardless payment IDs always start with `PM`).

danworth: Thanks, changed!

PR #506, opened 2019-06-14
Pp 5248 renaming transaction to payment

danworth:
## WHAT YOU DID
This continue to rename `transaction` to `payment` focusing on the API layer in a coordinate effort with @alexbishop1 who is focusing on associated changes within consuming projects.

This does not rename occurrences within `GoCardlessPayment` and `DirectDebitEvent` since they will be removed shortly. 

This includes a commit to disable pact tests whilst we transition.


src/main/java/uk/gov/pay/directdebit/mandate/api/DirectDebitInfoFrontendResponse.java
```diff
@@ -21,96 +21,11 @@
 @JsonFormat(shape = JsonFormat.Shape.OBJECT)
```
danworth: I restructured this class to bring fields to the top and static classes to the bottom to make it easier to read what this class as a response looks like.

PR #510, opened 2019-06-17
PP-5298 Add ProviderID and MandateID to Create Payment response

RoryMalcolm:
-  Adds logic to return the provider ID and mandate ID in the
   response object upon the creation of a payment

-  Currently does not return provider ID but will do so when
   the provider ID logic is implemented

-  Tests have been updated to ensure that the mandate ID is
   correctly returned

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/payments/api/CollectPaymentResponse.java
+++ b/src/main/java/uk/gov/pay/directdebit/payments/api/CollectPaymentResponse.java
@@ -4,15 +4,20 @@ import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
 import uk.gov.pay.commons.api.json.ApiResponseDateTimeSerializer;
+import uk.gov.pay.directdebit.mandate.model.subtype.MandateExternalId;
 import uk.gov.pay.directdebit.payments.model.Payment;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/directdebit/payments/api/CollectPaymentResponse.java
```diff
@@ -26,6 +28,12 @@
     @JsonProperty
     private Long amount;
 
+    @JsonProperty("mandate_id")
+    private String mandateId;
+    
+    @JsonProperty("provider_id")
+    private String providerId;
```
danworth: There is a `PaymentProviderPaymentId` class to represent this, please can you use that instead of String?

src/main/java/uk/gov/pay/directdebit/payments/api/CollectPaymentResponse.java
```diff
@@ -26,6 +28,12 @@
     @JsonProperty
     private Long amount;
 
+    @JsonProperty("mandate_id")
+    private String mandateId;
```
danworth: If this is the mandate external Id then there is a `MandateExternalId` class to represent it.

src/test/java/uk/gov/pay/directdebit/payments/resources/PaymentResourceIT.java
```diff
@@ -40,8 +40,7 @@
 import static io.restassured.RestAssured.given;
 import static io.restassured.http.ContentType.JSON;
 import static javax.ws.rs.core.Response.Status.OK;
-import static org.hamcrest.Matchers.hasSize;
-import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.*;
```
danworth: You might need to update your ide to prevent it adding wildcard imports, some details here, we don't use wildcard imports.
https://gds-way.cloudapps.digital/manuals/programming-languages/java.html#imports

PR #517, opened 2019-06-17
PP-5297: Return payment provider in create mandate response

oswaldquek:


src/main/java/uk/gov/pay/directdebit/mandate/api/CreateMandateResponse.java
```diff
@@ -45,6 +46,9 @@
 
     @JsonProperty("description")
     private String description;
+    
+    @JsonProperty("payment_provider")
+    private String paymentProvider = PaymentProvider.GOCARDLESS.toString().toLowerCase();
```
stephencdaly: Can't we get the actual value from the gateway account when we create this response?

oswaldquek: We could but we don't really have to. it's TDD and as long as the test passes we're good. There isn't going to be another direct debit payment provider anytime soon.

alexbishop1: We do have another provider: sandbox.

oswaldquek: sorry, you're right. i've fixed this now.

PR #519, opened 2019-06-17
PP-5248 Payment.getPaymentProviderPaymentId(): return Optional

alexbishop1:
Make the `Payment.getPaymentProviderPaymentId()` method on `Payment` return an `Optional<PaymentProviderPaymentId>` because there may not always be one.


src/main/java/uk/gov/pay/directdebit/payments/model/Payment.java
```diff
@@ -104,8 +104,8 @@ public void setMandate(Mandate mandate) {
         this.mandate = mandate;
     }
 
-    public PaymentProviderPaymentId getProviderId() {
-        return providerId;
+    public Optional<PaymentProviderPaymentId> getProviderId() {
```
danworth: `Optional.of(👍 )`

PR #527, opened 2019-06-19
PP-5248 Make Mandate Immutable

danworth:
For consistency with `Payment` which has been made immutable this makes `Mandate`
immutable. Modifications are now performed by taking a copy and using a builder.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/model/Mandate.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/model/Mandate.java
@@ -10,27 +10,27 @@ import java.util.Optional;
 
 public class Mandate {
     private final String description;
-    private Long id;
+    private final Long id;
     private final MandateExternalId externalId;
-    private MandateState state;
+    private final MandateState state;
     private final GatewayAccount gatewayAccount;
     private final String returnUrl;
-    private MandateBankStatementReference mandateBankStatementReference;
+    private final MandateBankStatementReference mandateBankStatementReference;
     private final String serviceReference;
     private final ZonedDateTime createdDate;
-    private Payer payer;
-    private PaymentProviderMandateId paymentProviderMandateId;
+    private final Payer payer;
+    private final PaymentProviderMandateId paymentProviderMandateId;
 
     private Mandate(MandateBuilder builder) {
         this.id = builder.id;
-        this.gatewayAccount = builder.gatewayAccount;
-        this.externalId = builder.externalId;
-        this.mandateBankStatementReference = builder.mandateBankStatementReference;
-        this.serviceReference = builder.serviceReference;
-        this.state = builder.state;
-        this.returnUrl = builder.returnUrl;
-        this.createdDate = builder.createdDate;
+        this.gatewayAccount = Objects.requireNonNull(builder.gatewayAccount);
+        this.externalId = Objects.requireNonNull(builder.externalId);
+        this.serviceReference = Objects.requireNonNull(builder.serviceReference);
+        this.state = Objects.requireNonNull(builder.state);
+        this.returnUrl = Objects.requireNonNull(builder.returnUrl);
+        this.createdDate = Objects.requireNonNull(builder.createdDate);
         this.payer = builder.payer;
+        this.mandateBankStatementReference = builder.mandateBankStatementReference;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/directdebit/mandate/model/Mandate.java
```diff
@@ -10,16 +10,16 @@
...
+    private final MandateBankStatementReference mandateBankStatementReference;
     private final String serviceReference;
     private final ZonedDateTime createdDate;
-    private Payer payer;
-    private PaymentProviderMandateId paymentProviderMandateId;
+    private final Payer payer;
+    private final PaymentProviderMandateId paymentProviderMandateId;
 
     private Mandate(MandateBuilder builder) {
         this.id = builder.id;
```
alexbishop1: Can we add some `Objects.requireNonNull(…)` magic to fields that always have to be present?

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateStateUpdateService.java
```diff
@@ -50,6 +51,7 @@ void mandateCreatedFor(Mandate mandate) {
     Mandate confirmedOnDemandDirectDebitDetailsFor(Mandate mandate) {
         Mandate updatedMandate = confirmedDetailsFor(mandate);
         userNotificationService.sendOnDemandMandateCreatedEmailFor(updatedMandate);
+        directDebitEventService.registerDirectDebitConfirmedEventFor(updatedMandate);
```
alexbishop1: Why did this have to be added?

src/test/java/uk/gov/pay/directdebit/mandate/services/MandateStateUpdateServiceTest.java
```diff
@@ -148,26 +172,13 @@ public void mandatePendingFor_shouldRegisterAMandatePendingEvent() {
...
+                .build();
 
         service.mandatePendingFor(mandate);
 
-        verify(mockedDirectDebitEventService).registerMandatePendingEventFor(mandate);
-        assertThat(mandate.getState(), is(PENDING));
-    }
-
-    @Test
-    public void shouldThrowWhenSubmittingTheSameDetailsSeveralTimes_andCreateOnlyOneEvent_whenConfirmingDirectDebitDetails() {
```
alexbishop1: Why was this test removed?

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateStateUpdateService.java
danworth: The only place this method to register a DDEvent when confirming a mandate was in the test further down (which didn't work as expected and therefor never alerted us to the fact that this method was never being called when confirming a mandate). Therefor I believe we should be registering a confirm mandate event when we confirm a mandate, so I've added it here.

src/test/java/uk/gov/pay/directdebit/mandate/services/MandateStateUpdateServiceTest.java
danworth: The test never worked as expected since lines 169 and 170 we're never being invoked; execution stopped after checking the thrown exception as a result of line 167.

Now that we don't mutate mandates, it doesn't make sense to invoke `service.confirmedOnDemandDirectDebitDetailsFor(onDemandMandate);` with the same mandate object twice, to confirm it will throw an exception, since the immutable mandate will keep on passing. There is a test that confirms we cannot confirm a mandate that is in the state of `DIRECT_DEBIT_DETAILS_CONFIRMED`.

danworth: I have added an assertion to ensure `registerDirectDebitConfirmedEventFor` is being called when confirming a mandate within test `shouldUpdateMandateStateAndRegisterEventWhenConfirmingDirectDebitDetails_andSendEmail`

src/main/java/uk/gov/pay/directdebit/mandate/model/Mandate.java
danworth: yes, great idea, will do.

PR #536, opened 2019-06-21
PP-5299: Provider state for adding a specific mandate reference

oswaldquek:
This fixes https://github.com/alphagov/pay-publicapi/pull/549/

### DIFF
--- a/src/test/java/uk/gov/pay/directdebit/pact/PublicApiContractTest.java
+++ b/src/test/java/uk/gov/pay/directdebit/pact/PublicApiContractTest.java
@@ -16,7 +16,9 @@ import uk.gov.pay.directdebit.common.util.RandomIdGenerator;
 import uk.gov.pay.directdebit.junit.DropwizardAppWithPostgresRule;
 import uk.gov.pay.directdebit.mandate.dao.MandateDao;
 import uk.gov.pay.directdebit.mandate.fixtures.MandateFixture;
+import uk.gov.pay.directdebit.mandate.model.GoCardlessMandateId;
 import uk.gov.pay.directdebit.mandate.model.Mandate;
+import uk.gov.pay.directdebit.mandate.model.MandateBankStatementReference;
 import uk.gov.pay.directdebit.mandate.model.subtype.MandateExternalId;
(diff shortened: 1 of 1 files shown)


src/test/java/uk/gov/pay/directdebit/pact/PublicApiContractTest.java
```diff
@@ -69,6 +73,18 @@ public void aGatewayAccountWithExternalIdAndAMandateWithExternalIdExist(Map<Stri
...
+    @State("a gateway account with external id and a confirmed mandate exists")
+    public void aGatewayAccountWithExternalIdAndAConfirmedMandateExists(Map<String, String> params) {
+        aGatewayAccountWithExternalIdAndAMandateWithExternalIdExist(params);
+        MandateDao mandateDao = app.getApplication().getInstance(MandateDao.class);
+        Mandate mandate = mandateDao.findByExternalId(MandateExternalId.valueOf(params.get("mandate_id"))).get();
+        Mandate updatedMandate = fromMandate(mandate)
+                .withMandateBankStatementReference(MandateBankStatementReference.valueOf(params.get("bank_mandate_reference")))
+                .withPaymentProviderId(GoCardlessMandateId.valueOf(params.get("unique_identifier")))
+                .build();
+        mandateDao.updateReferenceAndPaymentProviderId(updatedMandate);
```
stephencdaly: Why do we need the DAO to update? Can't we just add these fields when we insert using the database fixture?

oswaldquek: It already adds the fields when inserting using the database fixture, but I want to get rid of that because the mandate bank ref and provider id aren't available on the creation of a mandate, so https://github.com/alphagov/pay-publicapi/blob/master/src/test/resources/pacts/publicapi-direct-debit-connector-create-mandate.json#L36 is essentially incorrect and I want to rectify that after this.

oswaldquek: The publicapi-direct-debit-connector-create-mandate is incorrect and mandate_reference should be null.

PR #541, opened 2019-06-24
PP-5299: Provider id and mandate reference should be null

oswaldquek:
These should be null on mandate creation, when the payment provider is
GoCardless.

Also removed
MandateResourceIT.confirm_shouldCreateAMandateWithoutPayment_ifMandateIsOnDemand
as it's covered by confirm_shouldCreateACustomerBankAccountMandate.

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/services/MandateService.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/services/MandateService.java
@@ -1,7 +1,6 @@
 package uk.gov.pay.directdebit.mandate.services;
 
 import com.google.common.collect.ImmutableMap;
-import org.apache.commons.lang3.RandomStringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.directdebit.app.config.DirectDebitConfig;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/directdebit/mandate/services/MandateService.java
```diff
@@ -84,17 +87,13 @@ public Mandate createMandate(CreateMandateRequest createRequest, String accountE
...
                     }
 
-                    // TODO:
-                    // when we introduce GoCardless gateway accounts to work with create mandate,
-                    // then modify appropriate mandate reference values
-                    MandateBankStatementReference mandateReference = MandateBankStatementReference.valueOf(
-                            PaymentProvider.SANDBOX.equals(gatewayAccount.getPaymentProvider()) ?
-                                    RandomStringUtils.randomAlphanumeric(18) : "gocardless-default");
+                    var mandateReference = SANDBOX.equals(gatewayAccount.getPaymentProvider()) ?
+                            MandateBankStatementReference.valueOf(randomAlphanumeric(18)) : null;
```
stephencdaly: Shouldn't this be set when the mandate is confirmed for a consistent experience between Go Cardless and Sandbox accounts?

PR #542, opened 2019-06-24
PP-5299: Return provider id and mandate ref on mandate confirmation

oswaldquek:
This should fix https://github.com/alphagov/pay-publicapi/pull/553


src/main/java/uk/gov/pay/directdebit/mandate/api/GetMandateResponse.java
```diff
@@ -16,38 +18,38 @@
 @JsonInclude(Include.NON_NULL)
 @JsonFormat(shape = JsonFormat.Shape.OBJECT)
 public class GetMandateResponse {
+
+    @JsonProperty("provider_id")
+    private final String paymentProviderId;
```
stephencdaly: I think this should use the PaymentProviderPaymentId type with a `ToStringSerializer`

oswaldquek: Thanks for reviewing. I've fixed this.

PR #544, opened 2019-06-24
PP-5299: The mandate bank statement reference is optional

oswaldquek:
## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
+++ b/src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
@@ -15,6 +15,7 @@ import javax.inject.Inject;
 import java.math.BigDecimal;
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
+import java.util.Map;
 import java.util.Optional;
 
 public class UserNotificationService {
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
```diff
@@ -42,7 +42,7 @@ public UserNotificationService(AdminUsersClient adminUsersClient, DirectDebitCon
     public void sendMandateFailedEmailFor(Mandate mandate) {
         adminUsersClient.sendEmail(EmailTemplate.MANDATE_FAILED, mandate,
                 ImmutableMap.of(
-                        MANDATE_REFERENCE_KEY, mandate.getMandateBankStatementReference().toString(),
+                        MANDATE_REFERENCE_KEY, mandate.getMandateBankStatementReference().map(x -> x.toString()).orElse(""),
```
stephencdaly: Will this show up in the email in an acceptable way? Or would it be better to give some other string instead which would be clearer in the email?

oswaldquek: I don't have the answer to this question but given it was possible the mandate bank statement ref was populated by a randomly generated value by dd-connector I don't think this is any worse than it was.

stephencdaly: Would be good to check the email template I reckon, it'd be quicker to fix now if necessary than it coming back to bite us later

PR #545, opened 2019-06-24
PP-5299: Return dummy granular info

oswaldquek:
As agreed at the story kick off we've agreed the details should be 'example
details' for now.

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation


src/main/java/uk/gov/pay/directdebit/mandate/api/CreateMandateResponse.java
```diff
@@ -61,13 +36,7 @@ public CreateMandateResponse(MandateExternalId mandateId,
...
-        this.dataLinks = dataLinks;
-        this.mandateId = mandateId;
-        this.returnUrl = returnUrl;
-        this.createdDate = createdDate;
-        this.state = state;
-        this.serviceReference = serviceReference;
-        this.mandateReference = mandateReference;
+        super(mandateId, returnUrl, dataLinks, state, serviceReference, mandateReference, createdDate);
         this.description = description;
         this.paymentProvider = paymentProvider;
```
stephencdaly: Why isn't there a description or payment provider on the get mandate response? 

Is there any need to have separate models? The create mandate response is the same as the get mandate response but with some fields not returned because they haven't been set on the mandate yet right?

oswaldquek: I don't know why there isn't a description 🤷‍♂ as this is code I haven't changed. I can certainly add description as part of this PR though, seeing that it should get returned in the get mandate response as shown in the DD API spec: https://docs.google.com/document/d/1P30yiPM9rW8S6B-oPIR9VNrGrG1kwRxIEnV__T2CO3s/edit#.

The DD API spec doesn't specify a payment provider though...

I think it makes sense for the Get and Create responses to be the same though, I'll amend this.

oswaldquek: I've updated this.

PR #552, opened 2019-06-26
PP-5299: Don't return next_url and next_url_post links

oswaldquek:
Do not merge until https://github.com/alphagov/pay-publicapi/pull/567 has been deployed through to production.

### DIFF
--- a/src/test/java/uk/gov/pay/directdebit/util/ResponseDoesNotContainLinkMatcher.java
+++ b/src/test/java/uk/gov/pay/directdebit/util/ResponseDoesNotContainLinkMatcher.java
@@ -1,0 +1,34 @@
+package uk.gov.pay.directdebit.util;
+
+import org.hamcrest.Description;
+import org.hamcrest.TypeSafeMatcher;
+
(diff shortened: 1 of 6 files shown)


src/test/java/uk/gov/pay/directdebit/util/ResponseDoesNotContainLinkMatcher.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class ResponseDoesNotContainLinkMatcher extends TypeSafeMatcher<List<Map<String, Object>>> {
+
+    private final String rel;
+
+    private ResponseDoesNotContainLinkMatcher(String rel) {
```
alexbishop1: I wonder if a `ResponseContainsLinkMatcher` (whose result could be inverted) might be more flexible for the future but given we only need the _not_ case for now, let’s leave it as it is.

PR #563, opened 2019-06-28
PP-5371 Respect organisation when retrieving mandates/payments by GC ID

alexbishop1:
GoCardless mandate/payment IDs are not unique across different GoCardlessaccounts, so when looking up a mandate/payment by its GoCardless ID, also check the GoCardless organisation ID (which we store alongside the gateway account).

This is simple enough to do because we only ever use the GoCardless ID to look up a mandate/payment if we’re processing a webhook event and events sent by GoCardless also contain the relevant organisation ID.

This PR consists of multiple logical commits: it may be easier to review commit by commit.


src/main/java/uk/gov/pay/directdebit/gatewayaccounts/model/PaymentProviderServiceId.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.pay.directdebit.gatewayaccounts.model;
+
+import uk.gov.pay.commons.model.WrappedStringValue;
+
+/**
+ * The ID used by a payment provider to refer to a service that takes
+ * payments from users
+ */
+public abstract class PaymentProviderServiceId extends WrappedStringValue {
```
oswaldquek: The only class that extends this is GoCardlessOrganisationId at the moment. Is the idea that there will be one for Sandbox later?

alexbishop1: Probably but not until a bit later when we start filling out the sandbox functionality.

PR #572, opened 2019-07-02
Pp 5302 search by state and dates

oswaldquek:
## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
@@ -89,8 +89,8 @@ public class MandateSearchDao {
         params.getCaseInsensitivePartialSearchTermsOnMandate().forEach(e -> queryMap.put(e.getKey(), "%" + e.getValue() + "%"));
         params.getCaseInsensitivePartialSearchTermsOnPayer().forEach(e -> queryMap.put(e.getKey(), "%" + e.getValue() + "%"));
         params.getExactSearchTermsOnMandate().forEach(e -> queryMap.put(e.getKey(), e.getValue()));
-        params.getGreaterThanSearchTermsOnMandate().forEach(e -> queryMap.put(e.getKey(), e.getValue()));
-        params.getLessThanSearchTermsOnMandate().forEach(e -> queryMap.put(e.getKey(), e.getValue()));
+        params.getGreaterThanSearchTermsOnMandate().forEach(e -> queryMap.put(e.getKey() + "_from", e.getValue()));
+        params.getLessThanSearchTermsOnMandate().forEach(e -> queryMap.put(e.getKey() + "_to", e.getValue()));
         return queryMap;
     }
 
@@ -99,8 +99,8 @@ public class MandateSearchDao {
         params.getCaseInsensitivePartialSearchTermsOnMandate().forEach(e -> sb.append(format(" AND m.%s ILIKE :%s", e.getKey(), e.getKey())));
         params.getCaseInsensitivePartialSearchTermsOnPayer().forEach(e -> sb.append(format(" AND p.%s ILIKE :%s", e.getKey(), e.getKey())));
         params.getExactSearchTermsOnMandate().forEach(e -> sb.append(format(" AND m.%s = :%s", e.getKey(), e.getKey())));
-        params.getGreaterThanSearchTermsOnMandate().forEach(e -> sb.append(format(" AND m.%s > :%s", e.getKey(), e.getKey())));
-        params.getLessThanSearchTermsOnMandate().forEach(e -> sb.append(format(" AND m.%s < :%s", e.getKey(), e.getKey())));
+        params.getGreaterThanSearchTermsOnMandate().forEach(e -> sb.append(format(" AND m.%s > :%s", e.getKey(), e.getKey() + "_from")));
+        params.getLessThanSearchTermsOnMandate().forEach(e -> sb.append(format(" AND m.%s < :%s", e.getKey(), e.getKey() + "_to")));
         return sb.toString();
     }
 
--- a/src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
@@ -7,6 +7,7 @@ import uk.gov.pay.directdebit.mandate.model.Mandate;
 import uk.gov.pay.directdebit.mandate.params.MandateSearchParams;
 
 import javax.inject.Inject;
+import java.time.ZonedDateTime;
 import java.util.AbstractMap.SimpleEntry;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
```diff
@@ -86,13 +88,19 @@ public MandateSearchDao(Jdbi jdbi) {
...
+        params.getGreaterThanSearchTermsOnMandate().forEach(e -> queryMap.put(e.getKey() + "_from", e.getValue()));
+        params.getLessThanSearchTermsOnMandate().forEach(e -> queryMap.put(e.getKey() + "_to", e.getValue()));
         return queryMap;
     }
 
     private String generateQuery(Params params) {
         StringBuilder sb = new StringBuilder();
         params.getCaseInsensitivePartialSearchTermsOnMandate().forEach(e -> sb.append(format(" AND m.%s ILIKE :%s", e.getKey(), e.getKey())));
         params.getCaseInsensitivePartialSearchTermsOnPayer().forEach(e -> sb.append(format(" AND p.%s ILIKE :%s", e.getKey(), e.getKey())));
+        params.getExactSearchTermsOnMandate().forEach(e -> sb.append(format(" AND m.%s = :%s", e.getKey(), e.getKey())));
```
alexbishop1: If we’ve already got a `StringBuilder`, we don’t really need to use `String.format(…)`, we can just do something like this:

```java
sb.append(" AND m.").append(e.getKey()).append(" = :").append(e.getKey());
```

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
```diff
@@ -86,13 +88,19 @@ public MandateSearchDao(Jdbi jdbi) {
...
+        params.getGreaterThanSearchTermsOnMandate().forEach(e -> sb.append(format(" AND m.%s > :%s", e.getKey(), e.getKey() + "_from")));
```
alexbishop1: It took me more time than I care to admit to determine that this doesn’t introduce a risk of SQL injection (because `e.getKey()` comes from predefined constant strings defined by us rather than any remote input. I wish this was more obvious.

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
```diff
@@ -117,17 +125,29 @@ private Params(MandateSearchParams mandateSearchParams) {
...
+        List<SimpleEntry<String, ?>> getLessThanSearchTermsOnMandate() {
+            return List.of(toDate).stream().filter(e -> e.getValue() != null).collect(Collectors.toList());
+        }
+
+        List<SimpleEntry<String, ?>> getGreaterThanSearchTermsOnMandate() {
+            return List.of(fromDate).stream().filter(e -> e.getValue() != null).collect(Collectors.toList());
+        }
+
+        List<SimpleEntry<String, ?>> getExactSearchTermsOnMandate() {
+            return List.of(mandateState).stream().filter(e -> isNotBlank(e.getValue())).collect(Collectors.toList());
```
alexbishop1: Why is this one `isNotBlank(…)` while the other ones are `!= null`?

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
```diff
@@ -117,17 +125,29 @@ private Params(MandateSearchParams mandateSearchParams) {
...
+        }
+
```
alexbishop1: I know you’re just modifying what’s already there but I find this class quite hard to follow. I’m not saying change it now, but fundamentally I don’t see why it can’t be simplified to something like this:

```java
public List<Mandate> search(MandateSearchParams mandateSearchParams) {
	var sql = new StringBuilder();
	sql.append("SELECT DISTINCT … WHERE g.external_id = :gatewayAccountExternalId");

	var sqlParams = new HashMap<>();
	sqlParams.put("gatewayAccountExternalId", mandateSearchParams.getGatewayAccountExternalId());

	Optional.ofNullable(mandateSearchParams.getReference()).ifPresent(reference -> {
		sql.append(" AND m.reference ILIKE :reference");
		sqlParams.put("reference", reference);
	});

	Optional.ofNullable(mandateSearchParams.getFromDate()).ifPresent(fromDate -> {
		sql.append(" AND m.created_date_from > :created_date_from");
		sqlParams.put("created_date_from", fromDate);
	});

	// etc. etc.

    return jdbi.withHandle(handle -> {
        Query query = handle.createQuery(sql.toString());
        sqlParams.forEach(query::bind);
        return query.map(new MandateMapper()).list();
    });
}
```

It would be more verbose but I think it would be a lot easier to follow.

src/test/java/uk/gov/pay/directdebit/mandate/dao/SearchMandateDaoIT.java
```diff
@@ -53,59 +59,84 @@ public void setup() {
     @Test
     @Parameters({"REF1234", "ref1234", "f12"})
     public void searchByReference(String searchString) {
-        var mandateSearchParams = aMandateSearchParams().withReference(searchString).withGatewayAccountId("gateway-account-id");
-        List<Mandate> results = mandateSearchDao.search(mandateSearchParams);
+        var searchParams = aMandateSearchParams().withReference(searchString).withGatewayAccountId("gateway-account-id");
+        List<Mandate> results = mandateSearchDao.search(searchParams);
         assertThat(results.size()).isEqualTo(1);
-        assertThat(results.get(0)).isEqualTo(mandate1.toEntity());
+        assertThat(results.get(0).getId()).isEqualTo(mandate1.getId());
```
alexbishop1: Why is this now only comparing the ID rather than the entire mandate?

src/test/java/uk/gov/pay/directdebit/mandate/dao/SearchMandateDaoIT.java
```diff
@@ -53,59 +59,84 @@ public void setup() {
...
-        
+        var searchParams = aMandateSearchParams().withFromDate(now().minusHours(1)).withGatewayAccountId("gateway-account-id");
+        List<Mandate> results = mandateSearchDao.search(searchParams);
+        assertThat(results.size()).isEqualTo(1);
+        assertThat(results.get(0).getId()).isEqualTo(mandate1.getId());
+
+        searchParams = aMandateSearchParams().withFromDate(now().minusHours(7)).withGatewayAccountId("gateway-account-id");
+        results = mandateSearchDao.search(searchParams);
+        assertThat(results.size()).isEqualTo(2);
+        assertThat(results.stream().map(m -> m.getId()).collect(toList())).contains(mandate1.getId(), mandate2.getId());
```
alexbishop1: `assertThat(results.stream().map(Mandate::getId).collect(toSet()), is(List.of(mandate1.getId(), mandate2.getId()))` might be clearer and will also check the size.

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
oswaldquek: It was the verbosity I was trying to avoid! Hence my attempt to make this more elegant by returning lists and mapping over them. However I think you're right, it's getting quite hard to read and I will use the more verbose style you have suggested for readability.

src/test/java/uk/gov/pay/directdebit/mandate/dao/SearchMandateDaoIT.java
oswaldquek: For some reason the date returned from the postgres database was BST, while it was UTC when setting up the mandate fixture. This caused the object equality to fail. I'm not sure if this is a quirk of the postgres-in-docker database.

oswaldquek: it was actually me not setting the UTC timezone when inserting the mandate into the database.

PR #574, opened 2019-07-03
PP-5377 Disallow Payment Creation on Invalid Mandate State

RoryMalcolm:
- If a mandate has an invalid state and cannot take payments, currently the payment is accepted then an error occurs further down the line of the user journey.
- This fix changes that and does not allow payments to be created if a mandate has an invalid state, instead an error is returned from DD connector which will be handled in publicapi.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/payments/resources/PaymentResource.java
+++ b/src/main/java/uk/gov/pay/directdebit/payments/resources/PaymentResource.java
@@ -4,6 +4,7 @@ import com.codahale.metrics.annotation.Timed;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.directdebit.gatewayaccounts.model.GatewayAccount;
+import uk.gov.pay.directdebit.mandate.exception.MandateStateInvalidException;
 import uk.gov.pay.directdebit.payments.api.CollectPaymentRequest;
 import uk.gov.pay.directdebit.payments.api.CollectPaymentRequestValidator;
 import uk.gov.pay.directdebit.payments.api.PaymentResponse;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/directdebit/payments/resources/PaymentResource.java
```diff
@@ -57,6 +58,10 @@ public Response collectPaymentFromMandate(@PathParam("accountId") GatewayAccount
         LOGGER.info("Received collect payment from mandate request");
         collectPaymentRequestValidator.validate(collectPaymentRequestMap);
         Payment paymentToCollect = collectService.collect(gatewayAccount, CollectPaymentRequest.of(collectPaymentRequestMap));
+        if(!paymentService.isPaymentMandateStateValid(paymentToCollect)) {
```
stephencdaly: can you auto-format please

stephencdaly: this should be done inside the collect service I think. we definitely shouldn't be doing it after we collect the payment.

RoryMalcolm: Refactored

src/test/java/uk/gov/pay/directdebit/payments/resources/PaymentResourceIT.java
```diff
@@ -311,6 +311,35 @@ public void shouldRetrieveAPayment_fromPublicApiEndpoint() {
...
+                .build());
+
+        String requestPath = "/v1/api/accounts/{accountId}/charges/collect"
+                .replace("{accountId}", accountExternalId);
+
+        ValidatableResponse response = givenSetup()
+                .body(postBody)
+                .post(requestPath)
+                .then()
+                .statusCode(500);
```
stephencdaly: should probably check the error_identifier here.

RoryMalcolm: Done

PR #576, opened 2019-07-03
Pp 5302 mandate search dao refactor

oswaldquek:
## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
@@ -7,66 +7,13 @@ import uk.gov.pay.directdebit.mandate.model.Mandate;
 import uk.gov.pay.directdebit.mandate.params.MandateSearchParams;
 
 import javax.inject.Inject;
-import java.time.ZonedDateTime;
-import java.util.AbstractMap.SimpleEntry;
 import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
-import java.util.stream.Collectors;
-
-import static java.lang.String.format;
-import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
 public class MandateSearchDao {
     
-    private static final String MANDATE_GATEWAY_ACCOUNT_ID_COLUMN_NAME = "gateway_account_id";
-    private static final String MANDATE_SERVICE_REFERENCE_COLUMN_NAME = "service_reference";
-    private static final String MANDATE_STATE_COLUMN_NAME = "state";
-    private static final String MANDATE_REFERENCE_COLUMN_NAME = "mandate_reference";
-    private static final String MANDATE_CREATED_DATE_COLUMN_NAME = "created_date";
-    
-    private static final String PAYER_NAME_COLUMN_NAME = "name";
-    private static final String PAYER_EMAIL_COLUMN_NAME = "email";
-
     private final Jdbi jdbi;
-
-    private final String QUERY_STRING = "SELECT DISTINCT" +
-            "  m.id AS mandate_id," +
-            "  m.external_id AS mandate_external_id," +
-            format("  m.%s AS mandate_mandate_reference,", MANDATE_REFERENCE_COLUMN_NAME) +
-            format("  m.%s AS mandate_service_reference,", MANDATE_SERVICE_REFERENCE_COLUMN_NAME) +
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
```diff
@@ -7,159 +7,98 @@
...
-
     @Inject
     public MandateSearchDao(Jdbi jdbi) {
         this.jdbi = jdbi;
     }
 
     public List<Mandate> search(MandateSearchParams mandateSearchParams) {
-        Params params = new Params(mandateSearchParams);
-        String searchExtraFields = generateQuery(params);
+        var sql = new StringBuilder();
```
alexbishop1: The default capacity of a `StringBuilder` is 16 and I think it roughly doubles in size every time it reaches the limit. Since this string is always going to be quite large, we’re going to have a few resizes. Can we do `new StringBuilder(2048)` instead?

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
```diff
@@ -7,159 +7,98 @@
...
+                "  p.bank_account_number_last_two_digits AS payer_bank_account_number_last_two_digits," +
+                "  p.bank_account_requires_authorisation AS payer_bank_account_requires_authorisation," +
+                "  p.bank_account_number AS payer_bank_account_number," +
+                "  p.bank_account_sort_code AS payer_bank_account_sort_code," +
+                "  p.bank_name AS payer_bank_name," +
+                "  p.created_date AS payer_created_date" +
+                " FROM mandates m" +
+                "  JOIN gateway_accounts g ON g.id = m.gateway_account_id " +
+                "  LEFT JOIN payers p ON p.mandate_id = m.id " +
+                "WHERE g.external_id = :gatewayAccountExternalId ");
```
alexbishop1: Don’t need the final space here

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
```diff
@@ -7,159 +7,98 @@
...
+        //            "  ORDER BY p.id DESC OFFSET :offset LIMIT :limit";
+
+        var sqlParams = new HashMap<String, Object>();
+        sqlParams.put("gatewayAccountExternalId", mandateSearchParams.getGatewayAccountExternalId());
+
+        Optional.ofNullable(mandateSearchParams.getReference()).ifPresent(reference -> {
```
alexbishop1: Could we change `MandateSearchParams` so that all the getters for non-mandatory fields return `Optional`s? That would mean this could just be `mandateSearchParams.getReference().ifPresent(…)`.

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
```diff
@@ -7,159 +7,98 @@
...
+            sql.append(" AND m.service_reference ILIKE :serviceReference");
+            sqlParams.put("serviceReference", "%" + reference + "%");
```
alexbishop1: Could we use the same variable name for both the Java parameter and the SQL parameter everywhere (e.g. standardise on either `reference` or `serviceReference` but not use both)? I think it might make it a bit easier to follow.

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
```diff
@@ -7,159 +7,98 @@
...
+
+        var sqlParams = new HashMap<String, Object>();
+        sqlParams.put("gatewayAccountExternalId", mandateSearchParams.getGatewayAccountExternalId());
+
+        mandateSearchParams.getReference().filter(s -> !s.isBlank()).ifPresent(serviceReference -> {
+            sql.append(" AND m.service_reference ILIKE :serviceReference");
+            sqlParams.put("serviceReference", "%" + serviceReference + "%");
+        });
+        
+        mandateSearchParams.getMandateBankStatementReference().ifPresent(bankStatementRef -> {
```
alexbishop1: Does this want an `.filter(s -> !s.isBlank())` too?

oswaldquek: fixed

PR #579, opened 2019-07-04
Pp 5302 validation failures at resource level

oswaldquek:
A little controversially I've put validation within the MandateSearchParams.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
@@ -1,16 +1,14 @@
 package uk.gov.pay.directdebit.mandate.params;
 
-import com.google.common.collect.Range;
-import uk.gov.pay.directdebit.common.exception.BadRequestException;
+import uk.gov.pay.commons.validation.ValidDate;
 import uk.gov.pay.directdebit.mandate.model.MandateBankStatementReference;
 import uk.gov.pay.directdebit.mandate.model.MandateState;
 
+import javax.validation.constraints.Max;
+import javax.validation.constraints.Min;
+import javax.ws.rs.QueryParam;
 import java.time.ZonedDateTime;
-import java.time.format.DateTimeParseException;
 import java.util.Optional;
-
-import static java.lang.String.format;
-import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
 public class MandateSearchParams {
 
@@ -19,14 +17,25 @@ public class MandateSearchParams {
     private MandateBankStatementReference mandateBankStatementReference;
     private String name;
     private String email;
-    private ZonedDateTime fromDate;
-    private ZonedDateTime toDate;
-    private int page;
-    private int displaySize = 500;
-    private String gatewayAccountExternalId;
 
-    private MandateSearchParams() {
-    }
+    @QueryParam("from_date")
+    @ValidDate(message = "Invalid attribute value: from_date. Must be a valid date")
+    private String fromDate;
+
+    @QueryParam("to_date")
+    @ValidDate(message = "Invalid attribute value: to_date. Must be a valid date")
+    private String toDate;
+
+    @QueryParam("page")
+    @Min(value = 1, message = "Invalid attribute value: page. Must be greater than or equal to {value}")
+    private Integer page = 0;
+
+    @QueryParam("display_size")
+    @Min(value = 1, message = "Invalid attribute value: display_size. Must be greater than or equal to {value}")
+    @Max(value = 500, message = "Invalid attribute value: display_size. Must be less than or equal to {value}")
+    private Integer displaySize = 500;
+    
+    private String gatewayAccountExternalId;
 
     public Optional<String> getReference() {
         return Optional.ofNullable(reference);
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
```diff
@@ -101,18 +107,47 @@ public MandateSearchParams withToDate(ZonedDateTime toDate) {
         return this;
     }
 
-    public MandateSearchParams withPage(int page) {
-        this.page = page;
+    public MandateSearchParams withPage(Integer page) {
+        if (page != null && page < 1) {
+            throw new BadRequestException("Query param 'page' should be a non zero positive integer.");
```
alexbishop1: This message (and others) matches what we have for Direct Debit payments but is slightly different from what we do in card connector (the message has a lower-case “q” for “query”). I don’t know how much this matters.

src/main/java/uk/gov/pay/directdebit/mandate/resources/MandateResource.java
```diff
@@ -41,6 +45,24 @@ public MandateResource(MandateService mandateService, MandateQueryService mandat
...
+                                                @QueryParam("display_size") Integer displaySize,
+                                                @QueryParam("page") Integer page,
+                                                @QueryParam("from_date") String fromDate,
+                                                @QueryParam("to_date") String toDate) {
+        var mandateSearchParams = aMandateSearchParams()
+                .withToDate(toDate)
+                .withFromDate(fromDate)
+                .withDisplaySize(displaySize)
+                .withPage(page);
+        return List.of();
```
alexbishop1: I assume the real code is coming in a future PR!

src/test/java/uk/gov/pay/directdebit/mandate/resources/MandateSearchResourceIT.java
```diff
@@ -1,14 +1,75 @@
...
     
     @Test
-    public void searchByInvalidDisplaySize() {
+    @Parameters({
+            "to_date, 2018-14-05T15:00Z, Input to_date (2018-14-05T15:00Z) is wrong format.", 
+            "from_date, 2018-14-05T15:00Z, Input from_date (2018-14-05T15:00Z) is wrong format.", 
+            "page, -1, Query param 'page' should be a non zero positive integer.", 
+            "display_size, 0, Query param 'display_size' should be between 1 and 500.", 
+            "display_size, 501, Query param 'display_size' should be between 1 and 500."
+    })
```
alexbishop1: Nice use of parameterised tests

src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
alexbishop1: Is it possible to reuse the existing `NegativeSearchParamException` here?

src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
```diff
@@ -101,18 +107,47 @@ public MandateSearchParams withToDate(ZonedDateTime toDate) {
...
+        }
+        this.page = Optional.ofNullable(page).orElse(0);
         return this;
     }
 
-    public MandateSearchParams withDisplaySize(int displaySize) {
-        this.displaySize = displaySize;
+    public MandateSearchParams withDisplaySize(Integer displaySize) {
+        int size = Optional.ofNullable(displaySize).orElse(500);
+        if (!Range.closed(1, 500).contains(size)) {
```
alexbishop1: It seems that when searching for Direct Debit payments we don’t actually check if the value is greater than 500, which is surprising, but we probably want to make it consistent one way or the other

src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
```diff
@@ -101,18 +107,47 @@ public MandateSearchParams withToDate(ZonedDateTime toDate) {
...
+    public MandateSearchParams withToDate(String toDate) {
+        if (isNotBlank(toDate)) {
+            try {
+                withToDate(ZonedDateTime.parse(toDate));
+            } catch (DateTimeParseException e) {
+                throw new BadRequestException(format("Input to_date (%s) is wrong format.", toDate));
+            }
+        }
+        return this;
+    }
```
alexbishop1: These two methods could probably share common code in a private method

src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
oswaldquek: So I didn't use `NegativeSearchParamException` because it had a pre-defined message of `Query param '%s' should be a non zero positive integer` whereas I wanted `Query param 'display_size' should be between 1 and 500.` I don't know what the benefit of using `NegativeSearchParamException` is anyway - what's wrong with throwing a `BadRequestException`? The error message is going to get logged so from the point of view of support it makes no difference to me.

src/main/java/uk/gov/pay/directdebit/mandate/resources/MandateResource.java
oswaldquek: Yes :) It's TDD and I haven't written the test for a successful search yet.

src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
oswaldquek: true

oswaldquek:
@alexbishop1 I think it's possible to use `@Valid`. I'll try it out and see. I didn't do it originally because `@Valid` is usually annotated on objects that should be validated by jersey before it hits the resource method but we have to have individual `QueryParam`s on each search fields as a parameter to the resource method here.

oswaldquek:
@alexbishop1 good call with using `@Valid`. it actually nicely dealt with all of your comments.

PR #583, opened 2019-07-07
PP-5318 Update GoCardless mandate state based on all events

alexbishop1:
After receiving a webhook from GoCardless, update the state of each affected mandate by examining all of the events stored for that mandate, rather than the one that’s just been received. This means that we can handle webhook events from GoCardless for mandates being received out of order.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/events/dao/GoCardlessEventDao.java
+++ b/src/main/java/uk/gov/pay/directdebit/events/dao/GoCardlessEventDao.java
@@ -4,14 +4,21 @@ import org.jdbi.v3.sqlobject.config.RegisterArgumentFactory;
 import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
 import org.jdbi.v3.sqlobject.customizer.Bind;
 import org.jdbi.v3.sqlobject.customizer.BindBean;
+import org.jdbi.v3.sqlobject.customizer.BindList;
 import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
 import org.jdbi.v3.sqlobject.statement.SqlUpdate;
 import uk.gov.pay.directdebit.events.dao.mapper.GoCardlessEventMapper;
 import uk.gov.pay.directdebit.events.model.GoCardlessOrganisationIdArgumentFactory;
+import uk.gov.pay.directdebit.gatewayaccounts.model.GoCardlessOrganisationId;
+import uk.gov.pay.directdebit.mandate.model.GoCardlessMandateId;
 import uk.gov.pay.directdebit.mandate.model.GoCardlessMandateIdArgumentFactory;
 import uk.gov.pay.directdebit.payments.model.GoCardlessEvent;
 import uk.gov.pay.directdebit.payments.model.GoCardlessEventIdArgumentFactory;
 import uk.gov.pay.directdebit.payments.model.GoCardlessPaymentIdArgumentFactory;
+
+import java.util.Optional;
+import java.util.Set;
 
 @RegisterArgumentFactory(GoCardlessEventIdArgumentFactory.class)
 @RegisterArgumentFactory(GoCardlessMandateIdArgumentFactory.class)
@@ -65,6 +72,40 @@ public interface GoCardlessEventDao {
     @GetGeneratedKeys
     Long insert(@BindBean GoCardlessEvent goCardlessEvent);
 
-    @SqlUpdate("UPDATE gocardless_events t SET internal_event_id = :eventId WHERE t.id = :id")
+    @SqlUpdate("UPDATE gocardless_events SET internal_event_id = :eventId WHERE id = :id")
     int updateEventId(@Bind("id") Long id, @Bind("eventId") Long eventId);
+    
+    @SqlQuery("SELECT id, " +
+            "internal_event_id, " +
+            "event_id, " +
+            "action, " +
+            "created_at, " +
+            "details_cause, " +
+            "details_description, " +
+            "details_origin, " +
+            "details_reason_code, " +
+            "details_scheme, " +
+            "resource_type," +
+            "links_mandate, " +
+            "links_new_customer_bank_account, " +
+            "links_new_mandate, " +
+            "links_organisation, " +
+            "links_parent_event, " +
+            "links_payment, " +
+            "links_payout, " +
+            "links_previous_customer_bank_account, " +
+            "links_refund, " +
+            "links_subscription, " +
+            "json, " +
+            "event_id " +
+            "FROM gocardless_events " +
+            "WHERE links_mandate = :linksMandate " +
+            "AND links_organisation = :linksOrganisation " +
+            "AND action IN (<applicableActions>) " +
+            "ORDER BY created_at DESC " +
+            "LIMIT 1")
+    Optional<GoCardlessEvent> findLatestApplicableEventForMandate(@Bind("linksMandate") GoCardlessMandateId goCardlessMandateId,
+                                                                        @Bind("linksOrganisation") GoCardlessOrganisationId goCardlessOrganisationId,
+                                                                        @BindList("applicableActions") Set<String> applicableActions);
+
 }
--- a/src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
@@ -114,6 +114,13 @@ public interface MandateDao {
     @SqlUpdate("UPDATE mandates m SET state = :state WHERE m.id = :id")
     int updateState(@Bind("id") Long id, @Bind("state") MandateState mandateState);
 
+    @SqlUpdate("UPDATE mandates m SET state = :state FROM gateway_accounts g WHERE m.payment_provider_id = :providerMandateId " +
+            "AND g.id = m.gateway_account_id AND g.organisation = :providerServiceId AND g.payment_provider = :provider")
+    int updateStateByPaymentProviderMandateId(@Bind("provider") PaymentProvider paymentProvider,
+                                              @Bind("providerServiceId") PaymentProviderServiceId paymentProviderServiceId,
+                                              @Bind("providerMandateId") PaymentProviderMandateId paymentProviderMandateId,
+                                              @Bind("state") MandateState mandateState);
+
     @SqlUpdate("UPDATE mandates m SET mandate_reference = :mandateBankStatementReference, payment_provider_id = :paymentProviderMandateId WHERE m.id = :id")
     int updateReferenceAndPaymentProviderId(@BindBean Mandate mandate);
 }
--- a/src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateCalculator.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateCalculator.java
@@ -1,0 +1,45 @@
+package uk.gov.pay.directdebit.mandate.services.gocardless;
+
+import uk.gov.pay.directdebit.events.dao.GoCardlessEventDao;
+import uk.gov.pay.directdebit.gatewayaccounts.model.GoCardlessOrganisationId;
+import uk.gov.pay.directdebit.mandate.model.GoCardlessMandateId;
+import uk.gov.pay.directdebit.mandate.model.MandateState;
+import uk.gov.pay.directdebit.payments.model.GoCardlessEvent;
+
+import javax.inject.Inject;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import static uk.gov.pay.directdebit.mandate.model.MandateState.ACTIVE;
+import static uk.gov.pay.directdebit.mandate.model.MandateState.CANCELLED;
+import static uk.gov.pay.directdebit.mandate.model.MandateState.CREATED;
+import static uk.gov.pay.directdebit.mandate.model.MandateState.FAILED;
+import static uk.gov.pay.directdebit.mandate.model.MandateState.SUBMITTED;
+
+class GoCardlessMandateStateCalculator {
+
+    private final GoCardlessEventDao goCardlessEventDao;
+
+    private static final Map<String, MandateState> GOCARDLESS_ACTION_TO_MANDATE_STATE = Map.of(
+            "created", CREATED,
+            "submitted", SUBMITTED,
+            "active", ACTIVE,
+            "cancelled", CANCELLED,
+            "failed", FAILED
+    );
+
+    static final Set<String> GOCARDLESS_ACTIONS_THAT_CHANGE_STATE = GOCARDLESS_ACTION_TO_MANDATE_STATE.keySet();
+
+    @Inject
+    GoCardlessMandateStateCalculator(GoCardlessEventDao goCardlessEventDao) {
+        this.goCardlessEventDao = goCardlessEventDao;
+    }
+
+    Optional<MandateState> calculate(GoCardlessMandateId goCardlessMandateId, GoCardlessOrganisationId goCardlessOrganisationId) {
+        return goCardlessEventDao.findLatestApplicableEventForMandate(goCardlessMandateId, goCardlessOrganisationId, GOCARDLESS_ACTIONS_THAT_CHANGE_STATE)
+                .map(GoCardlessEvent::getAction)
+                .map(GOCARDLESS_ACTION_TO_MANDATE_STATE::get);
+    }
+
+}
--- a/src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateUpdater.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateUpdater.java
@@ -1,0 +1,42 @@
+package uk.gov.pay.directdebit.mandate.services.gocardless;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.directdebit.gatewayaccounts.model.GoCardlessOrganisationId;
+import uk.gov.pay.directdebit.mandate.dao.MandateDao;
+import uk.gov.pay.directdebit.mandate.model.GoCardlessMandateId;
+
+import javax.inject.Inject;
+
+import static java.lang.String.format;
+import static uk.gov.pay.directdebit.gatewayaccounts.model.PaymentProvider.GOCARDLESS;
+
+public class GoCardlessMandateStateUpdater {
+    
+    private static final Logger LOGGER = LoggerFactory.getLogger(GoCardlessMandateStateUpdater.class);
+
+    private final MandateDao mandateDao;
+    private final GoCardlessMandateStateCalculator goCardlessMandateStateCalculator;
+
+    @Inject
+    GoCardlessMandateStateUpdater(MandateDao mandateDao, GoCardlessMandateStateCalculator goCardlessMandateStateCalculator) {
+        this.mandateDao = mandateDao;
+        this.goCardlessMandateStateCalculator = goCardlessMandateStateCalculator;
+    }
+
+    public void updateState(GoCardlessMandateId goCardlessMandateId, GoCardlessOrganisationId goCardlessOrganisationId) {
+        goCardlessMandateStateCalculator.calculate(goCardlessMandateId, goCardlessOrganisationId)
+                .ifPresentOrElse(state -> {
+                    int updated = mandateDao.updateStateByPaymentProviderMandateId(GOCARDLESS, goCardlessOrganisationId, goCardlessMandateId, state);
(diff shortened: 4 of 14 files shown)


src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateCalculator.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private static final Map<String, MandateState> GOCARDLESS_ACTION_TO_MANDATE_STATE = Map.of(
+            "created", CREATED,
+            "submitted", SUBMITTED,
+            "active", ACTIVE,
+            "cancelled", CANCELLED,
+            "failed", FAILED
+    );
+
+    public static final List<String> GOCARDLESS_ACTIONS_THAT_CHANGE_STATE = List.copyOf(GOCARDLESS_ACTION_TO_MANDATE_STATE.keySet());
```
danworth: Wondering whether this collection of statuses should be part of `GoCardlessMandateDao` where it is used will `goCardlessEventDao.findLatestApplicableEventForMandate` ever be passed a different collection of states?

Also, this could be `Set<String>`

src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateCalculator.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Inject
+    public GoCardlessMandateStateCalculator(GoCardlessEventDao goCardlessEventDao) {
+        this.goCardlessEventDao = goCardlessEventDao;
+    }
+
+    public Optional<MandateState> calculate(GoCardlessMandateId goCardlessMandateId, GoCardlessOrganisationId goCardlessOrganisationId) {
+        return goCardlessEventDao.findLatestApplicableEventForMandate(goCardlessMandateId, goCardlessOrganisationId, GOCARDLESS_ACTIONS_THAT_CHANGE_STATE)
```
danworth: Made the comment on the Dao but perhaps we could use your new `GoCardlessMandateIdAndOrganisationId`?

src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateCalculator.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Map;
+import java.util.Optional;
+
+import static uk.gov.pay.directdebit.mandate.model.MandateState.ACTIVE;
+import static uk.gov.pay.directdebit.mandate.model.MandateState.CANCELLED;
+import static uk.gov.pay.directdebit.mandate.model.MandateState.CREATED;
+import static uk.gov.pay.directdebit.mandate.model.MandateState.FAILED;
+import static uk.gov.pay.directdebit.mandate.model.MandateState.SUBMITTED;
+
+public class GoCardlessMandateStateCalculator {
```
danworth: Can be package private (suggestion by Intelij). I like the idea of being strict about what can be accessed outside the package and keeping a tight public api for each package (as we've seen with Connector it would help should we ever want to carve it up and keeps things from directly talking to things they shouldn't and a tangled mess evolving over time). There's a few suggestions by intelij so if you agree with the principle I won't highlight each occurrence on classes and methods but they're highlighted by Intelij.

src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateCalculator.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private final GoCardlessEventDao goCardlessEventDao;
+
+    private static final Map<String, MandateState> GOCARDLESS_ACTION_TO_MANDATE_STATE = Map.of(
```
danworth: Wondering if this might work as a separate enum since its like a data type with each value having `GoCardlessAction` and `MandateAction`. I think this would mean it needn't be static on this class to make it accessible elsewhere and `GOCARDLESS_ACTIONS_THAT_CHANGE_STATE` just becomes the enum values and again needn't be declared static here. Also line 42 `.map(GOCARDLESS_ACTION_TO_MANDATE_STATE::get)` would become something like 
`.map(<EnumName>::getMandateState)` 🤷‍♂

src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateUpdater.java
```diff
@@ -0,0 +1,41 @@
...
+    @Inject
+    public GoCardlessMandateStateUpdater(MandateDao mandateDao, GoCardlessMandateStateCalculator goCardlessMandateStateCalculator) {
+        this.mandateDao = mandateDao;
+        this.goCardlessMandateStateCalculator = goCardlessMandateStateCalculator;
+    }
+
+    public void updateState(GoCardlessMandateId goCardlessMandateId, GoCardlessOrganisationId goCardlessOrganisationId) {
+        goCardlessMandateStateCalculator.calculate(goCardlessMandateId, goCardlessOrganisationId)
+                .ifPresent(state -> {
+                    int updated = mandateDao.updateStateByPaymentProviderMandateId(GOCARDLESS, goCardlessOrganisationId, goCardlessMandateId, state);
```
danworth: Could in-line 'update' but perhaps not as expressive e.g...
```
if (mandateDao.updateStateByPaymentProviderMandateId(GOCARDLESS, goCardlessOrganisationId, goCardlessMandateId, state) == 1) {
```

src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateUpdater.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Inject
+    public GoCardlessMandateStateUpdater(MandateDao mandateDao, GoCardlessMandateStateCalculator goCardlessMandateStateCalculator) {
+        this.mandateDao = mandateDao;
+        this.goCardlessMandateStateCalculator = goCardlessMandateStateCalculator;
+    }
+
+    public void updateState(GoCardlessMandateId goCardlessMandateId, GoCardlessOrganisationId goCardlessOrganisationId) {
+        goCardlessMandateStateCalculator.calculate(goCardlessMandateId, goCardlessOrganisationId)
+                .ifPresent(state -> {
```
danworth: Do we need to log anything if its empty?

src/main/java/uk/gov/pay/directdebit/payments/model/GoCardlessMandateIdAndOrganisationId.java
```diff
@@ -0,0 +1,45 @@
+package uk.gov.pay.directdebit.payments.model;
+
+import uk.gov.pay.directdebit.gatewayaccounts.model.GoCardlessOrganisationId;
+import uk.gov.pay.directdebit.mandate.model.GoCardlessMandateId;
+
+import java.util.Objects;
+
+public class GoCardlessMandateIdAndOrganisationId {
```
danworth: 👍

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
```diff
@@ -19,18 +28,35 @@
...
 
     public void handleEvents(List<GoCardlessEvent> events) {
         events.forEach(goCardlessService::storeEvent);
+
+        Map<GoCardlessResourceType, List<GoCardlessEvent>> eventsGroupedByResourceType = events.stream().collect(groupingBy(GoCardlessEvent::getResourceType));
+
+        List<GoCardlessEvent> eventsThatAffectMandates =  Optional.ofNullable(eventsGroupedByResourceType.get(GoCardlessResourceType.MANDATES))
+                .orElse(Collections.emptyList());
+
+        determineAffectedMandates(eventsThatAffectMandates)
```
danworth: Wondering to factor our the mandate processing. So the flow of this method becomes...
1. group events by type
2. process mandate events.
3. in future process other types (e.g. payments).
4. log some stuff

Something like...
```
Map<GoCardlessResourceType, List<GoCardlessEvent>> eventsGroupedByResourceType = events.stream().collect(groupingBy(GoCardlessEvent::getResourceType));
updateStatusForMandateEvents(eventsGroupedByResourceType.getOrDefault(GoCardlessResourceType.MANDATES, Collections.emptyList()));
```
```
private void updateStatusForMandateEvents(List<GoCardlessEvent> eventsThatAffectMandates) {
        determineAffectedMandates(eventsThatAffectMandates)
                .forEach(goCardlessMandateIdAndOrganisationId ->
                        goCardlessMandateStateUpdater.updateState(
                                goCardlessMandateIdAndOrganisationId.getGoCardlessMandateId(),
                                goCardlessMandateIdAndOrganisationId.getGoCardlessOrganisationId()
                        )
                );
    }
```

src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateUpdater.java
```diff
@@ -0,0 +1,41 @@
...
+    private final MandateDao mandateDao;
+    private final GoCardlessMandateStateCalculator goCardlessMandateStateCalculator;
+
+    @Inject
+    public GoCardlessMandateStateUpdater(MandateDao mandateDao, GoCardlessMandateStateCalculator goCardlessMandateStateCalculator) {
+        this.mandateDao = mandateDao;
+        this.goCardlessMandateStateCalculator = goCardlessMandateStateCalculator;
+    }
+
+    public void updateState(GoCardlessMandateId goCardlessMandateId, GoCardlessOrganisationId goCardlessOrganisationId) {
```
danworth: Can we use your new `GoCardlessMandateIdAndOrganisationId`

src/test/java/uk/gov/pay/directdebit/events/dao/GoCardlessEventDaoITest.java
```diff
@@ -4,26 +4,35 @@
...
+import static uk.gov.pay.directdebit.payments.fixtures.GoCardlessEventFixture.aGoCardlessEventFixture;
 
 @RunWith(DropwizardJUnitRunner.class)
 @DropwizardConfig(app = DirectDebitConnectorApp.class, config = "config/test-it-config.yaml")
 public class GoCardlessEventDaoITest {
 
     @DropwizardTestContext
     private TestContext testContext;
-
+
```
danworth: bit of erroneous space...

src/test/java/uk/gov/pay/directdebit/events/dao/GoCardlessEventDaoITest.java
```diff
@@ -63,4 +72,56 @@ public void shouldInsertAnEvent() {
         assertThat(goCardlessEvent.get("created_at"), is(Timestamp.from(goCardlessEventFixture.getCreatedAt().toInstant())));
     }
 
+    @Test
+    public void shouldFindLatestApplicableEventForMandate() {
+        GoCardlessEventFixture latestEvent = aGoCardlessEventFixture().withLinksMandate(GoCardlessMandateId.valueOf("Mandate ID we want"))
```
danworth: The values make this test nice to read and understand what should happen 👍

src/test/java/uk/gov/pay/directdebit/mandate/dao/MandateDaoIT.java
```diff
@@ -274,6 +274,44 @@ public void shouldNotUpdateAnythingIfTransactionDoesNotExist() {
         assertThat(numOfUpdatedMandates, is(0));
     }
 
+    @Test
+    public void shouldUpdateStateByProviderIdAndReturnNumberOfAffectedRows() {
+        GatewayAccountFixture goCardlessGatewayAccountFixture = GatewayAccountFixture.aGatewayAccountFixture()
```
danworth: The descriptive values really help to understand what all the set up if so  👍 👍 👍

src/main/java/uk/gov/pay/directdebit/events/dao/GoCardlessEventDao.java
```diff
@@ -65,6 +72,40 @@
...
+            "event_id " +
+            "FROM gocardless_events " +
+            "WHERE links_mandate = :linksMandate " +
+            "AND links_organisation = :linksOrganisation " +
+            "AND action IN (<applicableActions>) " +
+            "ORDER BY created_at DESC " +
+            "LIMIT 1")
+    Optional<GoCardlessEvent> findLatestApplicableEventForMandate(@Bind("linksMandate") GoCardlessMandateId goCardlessMandateId,
+                                                                        @Bind("linksOrganisation") GoCardlessOrganisationId goCardlessOrganisationId,
+                                                                        @BindList("applicableActions") List<String> applicableActions);
```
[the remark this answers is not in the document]
alexbishop1: The `Set` is a good idea; I’ll change that.

I’m not convinced about defining `applicableActions` here. It feels a bit too much like business logic to belong in a DAO. In addition, these are the applicable actions for updating the state. It’s possible we might want to retrieve other applicable actions for other things (e.g. actions that cause emails to be sent) in the future.

src/main/java/uk/gov/pay/directdebit/events/dao/GoCardlessEventDao.java
```diff
@@ -65,6 +72,40 @@
...
+            "json, " +
+            "event_id " +
+            "FROM gocardless_events " +
+            "WHERE links_mandate = :linksMandate " +
+            "AND links_organisation = :linksOrganisation " +
+            "AND action IN (<applicableActions>) " +
+            "ORDER BY created_at DESC " +
+            "LIMIT 1")
+    Optional<GoCardlessEvent> findLatestApplicableEventForMandate(@Bind("linksMandate") GoCardlessMandateId goCardlessMandateId,
+                                                                        @Bind("linksOrganisation") GoCardlessOrganisationId goCardlessOrganisationId,
```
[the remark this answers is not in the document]
alexbishop1: I didn’t really want to create `GoCardlessMandateIdAndOrganisationId` in the first place; I only did it because I needed an object that could encapsulate both for insertion into a `Set`

src/main/java/uk/gov/pay/directdebit/events/dao/GoCardlessEventDao.java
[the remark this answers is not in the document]
danworth: yep that's a good point about avoiding business logic in DAO.

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
```diff
@@ -114,6 +114,13 @@
     @SqlUpdate("UPDATE mandates m SET state = :state WHERE m.id = :id")
     int updateState(@Bind("id") Long id, @Bind("state") MandateState mandateState);
 
+    @SqlUpdate("UPDATE mandates m SET state = :state FROM gateway_accounts g WHERE m.payment_provider_id = :providerMandateId " +
+            "AND g.id = m.gateway_account_id AND g.organisation = :providerServiceId AND g.payment_provider = :provider")
+    int updateStateByPaymentProviderMandateId(@Bind("provider") PaymentProvider paymentProvider,
+                                              @Bind("providerServiceId") PaymentProviderServiceId paymentProviderServiceId,
```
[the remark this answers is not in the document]
alexbishop1: Yes, I suppose the combination of `GoCardlessMandateId` and `GoCardlessOrganisationId` could be considered to be a composite key of sorts: neither are particularly useful for retrieving things on their own.

On the “organisation”/“service” distinction: GoCardless calls it an organisation but we use it to discriminate between services, so I named the generic thing that. But perhaps that’s just trying to extrapolate a generalised concept where there is none.

src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateCalculator.java
alexbishop1: You’re repeating yourself 😀

src/main/java/uk/gov/pay/directdebit/events/dao/GoCardlessEventDao.java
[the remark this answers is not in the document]
danworth: Ah I see, I don't feel strongly, my suggestion was only based on `GoCardlessMandateIdAndOrganisationId` being available and it reducing the number of args to the method. I'm happy if you don't wish to use it here.

src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateCalculator.java
alexbishop1: OK, I’ll make it package-private

src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateCalculator.java
alexbishop1: Perhaps but I want to keep some separation from a GoCardless action and what we interpret that to mean. The same action could potentially mean different things (even represent different states) in different circumstances. I worry that using an enum might encourage the insidious idea that a GoCardless action is, to an extent, just a state.

(I think it would be desirable to use an enum instead of a string for representing the action though. The only reason I didn’t was because we already have one and it isn’t quite what we want. It could be repurposed though. I also worry that GoCardless might introduce other actions that would break the necessary string-to-enum conversion code we’d have. I might drop them an email to check they won’t do that.)

src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateUpdater.java
alexbishop1: Not especially keen on the condition of an `if` statement having a side-effect. I’m not too happy with how this block of code turned out in any case.

src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateUpdater.java
alexbishop1: Not sure. Do we need to know if we received an event that didn’t require us to update the state? I do want to give the logging here more thought though.

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
alexbishop1: That’s a good idea. I’ll do that.

src/test/java/uk/gov/pay/directdebit/events/dao/GoCardlessEventDaoITest.java
alexbishop1: 🤦‍♂

src/test/java/uk/gov/pay/directdebit/mandate/dao/MandateDaoIT.java
alexbishop1: I have my moments!

src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateUpdater.java
danworth: "Not especially keen on the condition of an if statement having a side-effect" hadn't thought of that. Good point!

src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateUpdater.java
danworth: I think that information could be useful, especially if we ever encounter a scenario where we should be updating the state but aren't for a particular event, at least there might be a log line telling us that such an event doesn't require an update (which I assume would be less frequent outcome than those that do?).

alexbishop1: OK, I’ll add something

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
alexbishop1: Also can’t believe I used:

```java
Optional.ofNullable(eventsGroupedByResourceType.get(GoCardlessResourceType.MANDATES)).orElse(Collections.emptyList());
```

… when I could have just used:
```java
eventsGroupedByResourceType.getOrDefault(GoCardlessResourceType.MANDATES, Collections.emptyList())
```

alexbishop1:
As discussed in person, I’ve made the more straightforward changes but we have greater plans for `GoCardlessMandateIdAndOrganisationId`, which will come in another PR!

PR #586, opened 2019-07-09
PP-5302 Modify PaymentViewReponse to be generic for re-use.

danworth:
- Rename `PaymentViewResponse` to `SearchResponse` and make it generic to permit
parameterisation of the contents of the `List` of results. This means it can be
reused for Mandates as well as Payments.
- Move into the common package.
- Refactor `PaymentSearchService` for clarity
- Rename `ViewPaginationBuilder` to `LinksForSearchResult` to hopefully better
explain what it does, also remove the notion of it being a builder because it isn't
and make as many fields `final` as possible.
- Note that within existing logic for `LinksForSearchResult.buildLinks()` it mutates
the page number within the `searchParams` it was provided to create the links
which is a little confusing. In future PR we hope to change this. This is why
`long originalPageNumber = validatedSearchParams.getPage();` has been added for now
within `PaymentSearchService.getPaymentSearchResponse()`.


src/main/java/uk/gov/pay/directdebit/common/model/SearchResponse.java
```diff
@@ -0,0 +1,72 @@
...
+        SearchResponse that = (SearchResponse) o;
+        return Objects.equals(gatewayExternalId, that.gatewayExternalId) &&
+                Objects.equals(total, that.total) &&
+                Objects.equals(page, that.page);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(gatewayExternalId, total, page);
+    }
```
alexbishop1: Should all the fields be in the `equals(…)` and `hashCode()` methods?

src/main/java/uk/gov/pay/directdebit/payments/model/LinksForSearchResult.java
```diff
@@ -12,20 +12,20 @@
 import static com.fasterxml.jackson.annotation.JsonInclude.Include;
 
 @JsonInclude(Include.NON_NULL)
-public class ViewPaginationBuilder {
+public class LinksForSearchResult {
```
alexbishop1: This is a much better name

src/main/java/uk/gov/pay/directdebit/payments/services/PaymentSearchService.java
```diff
@@ -22,31 +21,26 @@
...
         this.gatewayAccountDao = gatewayAccountDao;
     }
 
-    public PaymentViewResponse getPaymentSearchResponse(PaymentViewSearchParams searchParams) {
-        return gatewayAccountDao.findByExternalId(searchParams.getGatewayExternalId())
-                .map(gatewayAccount -> {
-                    PaymentViewSearchParams validatedSearchParams = paymentViewValidator.validateParams(searchParams);
-                    List<PaymentResponse> viewListResponse = Collections.emptyList();
-                    Long total = getTotal(validatedSearchParams);
-                    if (total > 0) {
```
alexbishop1: Why was this condition removed? Optimisation just not worth it?

src/main/java/uk/gov/pay/directdebit/common/model/SearchResponse.java
danworth: Fair point, I've regened the methods.

src/main/java/uk/gov/pay/directdebit/payments/services/PaymentSearchService.java
danworth: Going to add it back.

PR #587, opened 2019-07-09
PP-5318 MandateLookupKey for updating mandate state from GC events

alexbishop1:
GoCardless mandate IDs aren’t unique, so to retrieve a mandate from the database, we have use the mandate ID in combination with the GoCardless organisation ID. Sandbox mandate IDs don’t have this limitation and sandbox accounts don’t have organisation IDs.

Make things simpler for ourselves by introducing a `MandateUpdateService`, which acts as a façade over the `MandateDao` and takes a `MandateLookupKey` object. Both `GoCardlessMandateIdAndOrganisationId` (a value type that wraps a GoCardless mandate ID and organisation ID) and `SandboxMandateId` implement the `MandateLookupKey` marker interface. The
`MandateUpdateService` uses some fantastically non-polymorphic code to determine the precise runtime type of the `MandateLookupKey` it’s been passed and calls the appropriate method on the `MandateDao` depending on whether the lookup needs to take account of the organisation ID or not.

Make use of the new `MandateUpdateService` when updating the status of a mandate in response to a GoCardless webhook. This means we can pass it a `GoCardlessMandateIdAndOrganisationId` (rather than a `GoCardlessMandateId` and a `GoCardlessOrganisationId`).

Spread more widely, this pattern will allow us to remove `PaymentProviderServiceId`, which is the interface implemented by `GoCardlessOrganisationId` in order to maintain the fiction that all payment providers have this concept of organisation IDs.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
@@ -116,10 +116,16 @@ public interface MandateDao {
 
     @SqlUpdate("UPDATE mandates m SET state = :state FROM gateway_accounts g WHERE m.payment_provider_id = :providerMandateId " +
             "AND g.id = m.gateway_account_id AND g.organisation = :providerServiceId AND g.payment_provider = :provider")
-    int updateStateByPaymentProviderMandateId(@Bind("provider") PaymentProvider paymentProvider,
-                                              @Bind("providerServiceId") PaymentProviderServiceId paymentProviderServiceId,
-                                              @Bind("providerMandateId") PaymentProviderMandateId paymentProviderMandateId,
-                                              @Bind("state") MandateState mandateState);
+    int updateStateByPaymentProviderMandateIdAndOrganisationId(@Bind("provider") PaymentProvider paymentProvider,
+                                                               @Bind("providerServiceId") PaymentProviderServiceId paymentProviderServiceId,
+                                                               @Bind("providerMandateId") PaymentProviderMandateId paymentProviderMandateId,
+                                                               @Bind("state") MandateState mandateState);
+
+    @SqlUpdate("UPDATE mandates m SET state = :state FROM gateway_accounts g WHERE m.payment_provider_id = :providerMandateId " +
+            "AND g.id = m.gateway_account_id AND g.organisation IS NULL AND g.payment_provider = :provider")
+    int updateStateByPaymentProviderMandateIdAndOrganisationId(@Bind("provider") PaymentProvider paymentProvider,
+                                                               @Bind("providerMandateId") PaymentProviderMandateId paymentProviderMandateId,
+                                                               @Bind("state") MandateState mandateState);
 
     @SqlUpdate("UPDATE mandates m SET mandate_reference = :mandateBankStatementReference, payment_provider_id = :paymentProviderMandateId WHERE m.id = :id")
     int updateReferenceAndPaymentProviderId(@BindBean Mandate mandate);
--- a/src/main/java/uk/gov/pay/directdebit/mandate/services/MandateUpdateService.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/services/MandateUpdateService.java
@@ -1,0 +1,34 @@
+package uk.gov.pay.directdebit.mandate.services;
+
+import uk.gov.pay.directdebit.gatewayaccounts.model.PaymentProvider;
+import uk.gov.pay.directdebit.mandate.dao.MandateDao;
+import uk.gov.pay.directdebit.mandate.model.MandateLookupKey;
+import uk.gov.pay.directdebit.mandate.model.MandateState;
+import uk.gov.pay.directdebit.mandate.model.PaymentProviderMandateId;
+import uk.gov.pay.directdebit.mandate.model.SandboxMandateId;
+import uk.gov.pay.directdebit.payments.model.GoCardlessMandateIdAndOrganisationId;
+
+import javax.inject.Inject;
+
+public class MandateUpdateService {
+
+    private final MandateDao mandateDao;
+
+    @Inject
+    MandateUpdateService(MandateDao mandateDao) {
+        this.mandateDao = mandateDao;
+    }
+
+    public int updateStateByPaymentProviderMandateId(PaymentProvider paymentProvider, MandateLookupKey mandateLookupKey, MandateState mandateState) {
+        if (mandateLookupKey.getClass() == GoCardlessMandateIdAndOrganisationId.class) {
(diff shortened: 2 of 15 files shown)


src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
```diff
@@ -121,6 +121,12 @@ int updateStateByPaymentProviderMandateId(@Bind("provider") PaymentProvider paym
                                               @Bind("providerMandateId") PaymentProviderMandateId paymentProviderMandateId,
                                               @Bind("state") MandateState mandateState);
 
+    @SqlUpdate("UPDATE mandates m SET state = :state FROM gateway_accounts g WHERE m.payment_provider_id = :providerMandateId " +
+            "AND g.id = m.gateway_account_id AND g.organisation IS NULL AND g.payment_provider = :provider")
+    int updateStateByPaymentProviderMandateId(@Bind("provider") PaymentProvider paymentProvider,
```
danworth: Perhaps rename the existing `updateStateByPaymentProviderMandateId` to something like `updateStateByPaymentProviderMandateIdAndServiceId` to help make the distinction between these two methods?

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateUpdateService.java
```diff
@@ -0,0 +1,34 @@
+package uk.gov.pay.directdebit.mandate.services;
```
danworth: I like that we're separating out the functionality into smaller mandate services. Not especially related to this PR but there is `MandateService` which now appears to have creation logic, so that could be named something more expressive like `MandateCreateService`, then we'd have `MandateQueryService`, `MandateCreateService` and `MandateUpdateService`

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateUpdateService.java
```diff
@@ -0,0 +1,34 @@
...
+import uk.gov.pay.directdebit.payments.model.GoCardlessMandateIdAndOrganisationId;
+
+import javax.inject.Inject;
+
+public class MandateUpdateService {
+
+    private final MandateDao mandateDao;
+
+    @Inject
+    public MandateUpdateService(MandateDao mandateDao) {
```
danworth: Can be package private.

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateUpdateService.java
```diff
@@ -0,0 +1,34 @@
...
+        this.mandateDao = mandateDao;
+    }
+
+    public int updateStateByPaymentProviderMandateId(PaymentProvider paymentProvider, MandateLookupKey mandateLookupKey, MandateState mandateState) {
+        if (mandateLookupKey.getClass() == GoCardlessMandateIdAndOrganisationId.class) {
```
danworth: The switching logic, albeit non-polymorphic as you point out is nicely contained in this single method.

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
```diff
@@ -57,13 +57,7 @@ public void handleEvents(List<GoCardlessEvent> events) {
...
 
     private void updateStateForMandateEvents(List<GoCardlessEvent> eventsThatAffectMandates) {
-        determineAffectedMandates(eventsThatAffectMandates)
-                .forEach(goCardlessMandateIdAndOrganisationId ->
-                        goCardlessMandateStateUpdater.updateState(
-                                goCardlessMandateIdAndOrganisationId.getGoCardlessMandateId(),
-                                goCardlessMandateIdAndOrganisationId.getGoCardlessOrganisationId()
-                        )
-                );
+        determineAffectedMandates(eventsThatAffectMandates).forEach(goCardlessMandateStateUpdater::updateState);
```
danworth: 👍 looks cleaner

src/test/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateUpdaterTest.java
```diff
@@ -21,11 +22,11 @@
 @RunWith(MockitoJUnitRunner.class)
 public class GoCardlessMandateStateUpdaterTest {
 
-    private static final GoCardlessMandateId GOCARDLESS_MANDATE_ID = GoCardlessMandateId.valueOf("MD123");
-    private static final GoCardlessOrganisationId GOCARDLESS_ORGANISATION_ID = GoCardlessOrganisationId.valueOf("OR123");
+    private static final GoCardlessMandateIdAndOrganisationId GOCARDLESS_MANDATE_ID_AND_ORGANISATION_ID = new GoCardlessMandateIdAndOrganisationId(
```
danworth: might be a nice use of `var` since its clear what is being assigned.

src/test/java/uk/gov/pay/directdebit/mandate/services/MandateUpdateServiceTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+        int updated = mandateUpdateService.updateStateByPaymentProviderMandateId(SANDBOX, SANDBOX_MANDATE_ID, PENDING);
+
+        assertThat(updated, is(1));
+
+        verifyZeroInteractions(ignoreStubs(mockMandateDao));
+    }
+
+    @Test
+    public void updateStateByPaymentProviderMandateIdWithSandboxMandateIdReturnsUpdateCount() {
```
danworth: Nice, covers the three paths in the logic 👍

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
alexbishop1: Renamed to `updateStateByPaymentProviderMandateIdAndOrganisationId` (since it’s always going to be a GoCardless organisation ID)

src/main/java/uk/gov/pay/directdebit/mandate/services/MandateUpdateService.java
alexbishop1: Changed to package-private

src/test/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateUpdaterTest.java
alexbishop1: Unfortunately, we can’t use `var` here because it’s a field, not a local variable

PR #595, opened 2019-07-11
PP-5302 Make MandateSearchParams implement SearchParams

danworth:
- Add `buildQueryParamsString` to `MandateSearchParams` and add test to ensure
it builds the query string correctly.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.directdebit.mandate.params;
 
 import uk.gov.pay.commons.validation.ValidDate;
+import uk.gov.pay.directdebit.common.model.SearchParams;
 import uk.gov.pay.directdebit.mandate.model.MandateBankStatementReference;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
```diff
@@ -65,14 +85,65 @@
...
+
+        return query.toString();
+    }
+
+    private String appendQueryParam(String name, String value) {
+        return "&" + createQueryParamFor(name, value);
+    }
+
+    private String createQueryParamFor(String name, String value) {
+        return name + "=" + value;
```
alexbishop1: Should we be doing URL-encoding here?

danworth: We follow the same practice in Connector, build a string of parameters which is then passed to a uribuilder which calls an `encode` method with the query String it is provided.

danworth: I've created a bug ticket https://payments-platform.atlassian.net/browse/PP-5435 for the broader fix.

PR #598, opened 2019-07-11
PP-5302 Add logic for mandate search to MandateResource

danworth:
- Remove `GatwayAccountExternalId` from `MandateSearchParams` and `SearchParams` as it
is provided in the URL path and not the search params which means it cannot be added
to MandateSearchParams using `@QueryBean` deserialisation. Instead pass this value
to the necessary service and dao methods.
- Introduce `MandateResourceIT` tests for mandate search

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/services/MandateQueryService.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/services/MandateQueryService.java
@@ -10,23 +10,19 @@ import uk.gov.pay.directdebit.mandate.model.MandateState;
 import uk.gov.pay.directdebit.mandate.model.PaymentProviderMandateId;
 import uk.gov.pay.directdebit.mandate.model.SandboxMandateId;
 import uk.gov.pay.directdebit.mandate.model.subtype.MandateExternalId;
-import uk.gov.pay.directdebit.mandate.params.MandateSearchParams;
 import uk.gov.pay.directdebit.payments.model.GoCardlessMandateIdAndOrganisationId;
 
 import javax.inject.Inject;
 import java.time.ZonedDateTime;
-import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 
 public class MandateQueryService {
     private MandateDao mandateDao;
-    private MandateSearchDao mandateSearchDao;
 
     @Inject
-    public MandateQueryService(MandateDao mandateDao, MandateSearchDao mandateSearchDao) {
+    public MandateQueryService(MandateDao mandateDao) {
         this.mandateDao = mandateDao;
-        this.mandateSearchDao = mandateSearchDao;
     }
 
     public Mandate findByExternalId(MandateExternalId externalId) {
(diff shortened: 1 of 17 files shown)


src/main/java/uk/gov/pay/directdebit/mandate/services/MandateQueryService.java
```diff
@@ -61,4 +66,30 @@ public Mandate findById(Long mandateId) {
     public List<Mandate> findAllMandatesBySetOfStatesAndMaxCreationTime(Set<MandateState> states, ZonedDateTime cutOffTime) {
         return mandateDao.findAllMandatesBySetOfStatesAndMaxCreationTime(states, cutOffTime);
     }
+    
+    public MandateSearchResults search(MandateSearchParams params, String gatewayAccountExternalId) {
```
alexbishop1: Maybe add a unit test for this?

alexbishop1: I’d actually be tempted to put this in its own class rather than `MandateQueryService` (`MandateSearchService`?) so that each service is quite tightly focused. It feels like a search is distinct enough from a query (for a single mandate, up until this point) to be its own thing.

danworth: Added 👍

danworth: Ok done 👍

src/test/java/uk/gov/pay/directdebit/mandate/services/MandateSearchServiceTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Test
+    public void shouldReturnOneMandateForPageAndTotalOfTwoMatchingMandates() {
+        var params = MandateSearchParams.MandateSearchParamsBuilder.aMandateSearchParams()
+                .withName("aName")
+                .build();
+
+        var mandateListToReturn = List.of(
+                aMandateFixture().withServiceReference("expectedReference").toEntity()
+        );
```
alexbishop1: It’s a little confusing that there are ostensibly two results but this list only has one

src/test/java/uk/gov/pay/directdebit/mandate/services/MandateSearchServiceTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private MandateSearchService mandateSearchService;
+
+    @Before
+    public void setUp() {
+        mandateSearchService = new MandateSearchService(mockMandateSearchDao);
+    }
+
+    @Test
+    public void shouldReturnOneMandateForPageAndTotalOfTwoMatchingMandates() {
```
alexbishop1: Could you also add a test for when there’s no results? Check the other side of that ternary operator!

src/test/java/uk/gov/pay/directdebit/mandate/services/MandateSearchServiceTest.java
danworth: I've tried to address that in the method names with `getTotalMatchingMandates` and `getMandatesForRequestedPage`. I deliberately used different values because it's reasonable that the total count is greater than the mandates required just for the requested page.

PR #599, opened 2019-07-11
PP-5431 Add payment status updater that reads GoCardless events

alexbishop1:
Add a payment status updater that works by querying for stored GoCardless webhook events. It always uses the latest dated (not received/stored) event that has a relevant action and doesn’t enforce any idea of valid state transitions.

Note that this isn’t actually hooked up to anything yet.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
@@ -120,16 +120,16 @@ public interface MandateDao {
 
     @SqlUpdate("UPDATE mandates m SET state = :state FROM gateway_accounts g WHERE m.payment_provider_id = :providerMandateId " +
             "AND g.id = m.gateway_account_id AND g.organisation = :goCardlessOrganisationId AND g.payment_provider = :provider")
-    int updateStateByPaymentProviderMandateIdAndOrganisationId(@Bind("provider") PaymentProvider paymentProvider,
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
```diff
@@ -120,16 +120,16 @@
 
     @SqlUpdate("UPDATE mandates m SET state = :state FROM gateway_accounts g WHERE m.payment_provider_id = :providerMandateId " +
             "AND g.id = m.gateway_account_id AND g.organisation = :goCardlessOrganisationId AND g.payment_provider = :provider")
-    int updateStateByPaymentProviderMandateIdAndOrganisationId(@Bind("provider") PaymentProvider paymentProvider,
-                                                               @Bind("goCardlessOrganisationId") GoCardlessOrganisationId goCardlessOrganisationId,
-                                                               @Bind("providerMandateId") PaymentProviderMandateId paymentProviderMandateId,
-                                                               @Bind("state") MandateState mandateState);
+    int updateStateByPaymentProviderMandateId(@Bind("provider") PaymentProvider paymentProvider,
```
danworth: Could this method name include something about the organisation id to help differentiate it from the other one?

alexbishop1: I intended for them to have different names but I was cursed by not paying enough attention to an IDE rename that went too far! I’ve updated the name.

PR #601, opened 2019-07-12
PP-5302 Fix date assertion in test

danworth:
I wrote the test MandateSearchParamsTest.shouldCreateQueryWithAllParams()
which would only ever pass on the 11 July 2019. This fixes the assertion to pass
on any day, please forgive me.


danworth:
Let's never speak of this again

PR #602, opened 2019-07-12
PP-5431 Integration test for out-of-order GoCardless payment events

alexbishop1:
Because it’s always good to know things work.

### DIFF
--- a/src/test/java/uk/gov/pay/directdebit/webhook/gocardless/resources/WebhookGoCardlessResourceOutOfOrderPaymentEventsIT.java
+++ b/src/test/java/uk/gov/pay/directdebit/webhook/gocardless/resources/WebhookGoCardlessResourceOutOfOrderPaymentEventsIT.java
@@ -1,0 +1,121 @@
+package uk.gov.pay.directdebit.webhook.gocardless.resources;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.tomakehurst.wiremock.junit.WireMockRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import uk.gov.pay.directdebit.DirectDebitConnectorApp;
+import uk.gov.pay.directdebit.gatewayaccounts.model.GoCardlessOrganisationId;
+import uk.gov.pay.directdebit.gatewayaccounts.model.PaymentProvider;
+import uk.gov.pay.directdebit.junit.DropwizardConfig;
+import uk.gov.pay.directdebit.junit.DropwizardJUnitRunner;
+import uk.gov.pay.directdebit.junit.DropwizardTestContext;
+import uk.gov.pay.directdebit.junit.TestContext;
+import uk.gov.pay.directdebit.mandate.fixtures.MandateFixture;
+import uk.gov.pay.directdebit.mandate.model.GoCardlessMandateId;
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/directdebit/webhook/gocardless/resources/WebhookGoCardlessResourceOutOfOrderPaymentEventsIT.java
```diff
@@ -0,0 +1,121 @@
...
+import static org.hamcrest.core.Is.is;
+
+@RunWith(DropwizardJUnitRunner.class)
+@DropwizardConfig(app = DirectDebitConnectorApp.class, config = "config/test-it-config.yaml")
+public class WebhookGoCardlessResourceOutOfOrderPaymentEventsIT {
+
+    private static final GoCardlessOrganisationId GOCARDLESS_ORGANISATION_ID = GoCardlessOrganisationId.valueOf("OR123");
+    private static final GoCardlessPaymentId GOCARDLESS_PAYMENT_ID = GoCardlessPaymentId.valueOf("PM123");
+    
+    private static final Map<String, Object> PAYMENT_PAID_OUT_FAILED_OUT_OF_ORDER_WEBHOOK = Map.of(
```
danworth: Dropwizard has a handy `fixture` helper which would allow you to write this as json in separate file and then do something like the following to create it as a string. I find it easier to read json as json and the fixture function makes it easy to stringify which is where this is ending up (or deserialise into a class when necessary). It also means that such fixtures can be re-used in other tests etc.

https://www.dropwizard.io/1.3.12/docs/manual/testing.html#fixtures
```
import static io.dropwizard.testing.FixtureHelpers.fixture;

String jsonPayload = fixture("path to your file");
```

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/resources/WebhookGoCardlessResourceOutOfOrderPaymentEventsIT.java
```diff
@@ -0,0 +1,121 @@
...
+    @DropwizardTestContext
+    private TestContext testContext;
+
+    @Rule
+    public WireMockRule wireMockRule = new WireMockRule(10112);
+
+    private ObjectMapper objectMapper = new ObjectMapper().configure(ORDER_MAP_ENTRIES_BY_KEYS, true);
+
+    @Test
+    public void handlesOutOfOrderWebhooksByUsingLatestTimestampedEvent() throws JsonProcessingException {
```
danworth: This tests that multiple events in the same request are processed correctly when they are out of order. Should there be another test (or included within this one) where an out of order event already exists in the db (which was what I was originally looking for before I realised that both events are in the request).

alexbishop1: We have something like that in `WebhookGoCardlessResourceOutOfOrderMandateEventsIT`. However, for payments, there’s only a couple of event actions that affect the state (for now), so it’s harder to do that.

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/resources/WebhookGoCardlessResourceOutOfOrderPaymentEventsIT.java
alexbishop1: I didn’t know about that and will bear it mind for the future.

PR #603, opened 2019-07-12
PP-5354 Add govukpay_events table

stephencdaly:

### DIFF
--- a/src/main/resources/migrations/00060_create_table_govukpay_events.sql
+++ b/src/main/resources/migrations/00060_create_table_govukpay_events.sql
@@ -1,0 +1,20 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_table-govukpay_events
+CREATE TABLE govukpay_events (
+    id BIGSERIAL PRIMARY KEY,
+    mandate_id BIGINT,
+    payment_id BIGINT,
+    event_date TIMESTAMP WITH TIME ZONE DEFAULT (now() AT TIME ZONE 'utc') NOT NULL,
+    resource_type VARCHAR(32),
+    event_type VARCHAR(255)
+);
+--rollback drop table govukpay_events;
+
+--changeset uk.gov.pay:add_govukpay_events_mandates_fk
+ALTER TABLE govukpay_events ADD CONSTRAINT govukpay_events_mandates_fk FOREIGN KEY (mandate_id) REFERENCES mandates (id);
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/00060_create_table_govukpay_events.sql
```diff
@@ -0,0 +1,20 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_table-govukpay_events
+CREATE TABLE govukpay_events (
+    id BIGSERIAL PRIMARY KEY,
+    mandate_id BIGINT,
+    payment_id BIGINT,
+    event_date TIMESTAMP WITH TIME ZONE DEFAULT (now() AT TIME ZONE 'utc') NOT NULL,
+    resource_type VARCHAR(255),
```
alexbishop1: 255 is probably a bit on the large size

src/main/resources/migrations/00060_create_table_govukpay_events.sql
```diff
@@ -0,0 +1,20 @@
...
+    event_type VARCHAR(255)
```
alexbishop1: This is the what happened bit, right? GoCardless call this an action. Maybe because it actually _is_ the event (conceptually), we could call it a name or something. But what’s here will do.

src/main/resources/migrations/00060_create_table_govukpay_events.sql
stephencdaly: 20?

src/main/resources/migrations/00060_create_table_govukpay_events.sql
stephencdaly: i thought i'd go with what ledger calls things for some level of consistency

alexbishop1: Ah, I didn’t know about ledger using the name. That’s cool.

src/main/resources/migrations/00060_create_table_govukpay_events.sql
alexbishop1: Maybe a little higher to give ourselves some breathing room. 32?

stephencdaly: ok

stephencdaly: updated

PR #604, opened 2019-07-12
PP-5302 Change mandate search to use external mandate states

danworth:
Mandate search should work with external states not internal. Note that external
states can map to multiple internal states and we only store the internal state
in the table so conversion is necessary before searching.

## WHAT YOU DID
As described above. Note the next change will be to permit a list of external states to search on rather than a single value (maybe in another story...)

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
@@ -10,7 +10,10 @@ import javax.validation.constraints.Min;
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.QueryParam;
 import java.time.ZonedDateTime;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 public class MandateSearchParams implements SearchParams {
 
@@ -28,7 +31,7 @@ public class MandateSearchParams implements SearchParams {
     private String serviceReference;
 
     @QueryParam(STATE)
-    private MandateState mandateState;
+    private String externalMandateState;
 
     @QueryParam(BANK_STATEMENT_REFERENCE)
     private MandateBankStatementReference mandateBankStatementReference;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
```diff
@@ -107,7 +110,13 @@ public Integer getPage() {
     public Integer getDisplaySize() {
         return displaySize;
     }
-    
+
+    public List<MandateState> getInternalStates() {
+        return Arrays.stream(MandateState.values())
+                .filter(state -> state.toExternal().getState().equalsIgnoreCase(this.externalMandateState))
```
alexbishop1: Not 100% sure we need this to be case-insensitive; it doesn’t look like it is for card connector

src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
```diff
@@ -107,7 +110,13 @@ public Integer getPage() {
...
+                .collect(Collectors.toList());
```
alexbishop1: `toUnmodifiableList()` 😀

src/test/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDaoIT.java
```diff
@@ -191,4 +184,25 @@ public void searchByMultipleParams() {
...
+                .insert(testContext.getJdbi());
+
+        var searchParams = aMandateSearchParams()
+                .withExternalMandateState("pending")
+                .build();
+
+        var results = mandateSearchDao.search(searchParams, gatewayAccountFixture.getExternalId());
+        assertThat(results).hasSize(2);
+        assertThat(results.get(0).getState().toExternal().getState()).isEqualTo("pending");
+        assertThat(results.get(1).getState().toExternal().getState()).isEqualTo("pending");
```
alexbishop1: Might be easier just to assert that we specifically get `mandate1` and `mandate2` — I think it would make the intent a bit clearer

src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
danworth: Done!

src/main/java/uk/gov/pay/directdebit/mandate/params/MandateSearchParams.java
danworth: Ah lovely!

src/test/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDaoIT.java
danworth: yes that is clearer!

PR #605, opened 2019-07-12
PP-5432 Read all webhook events when updating sandbox payment state

alexbishop1:
After processing a sandbox webhook, update the states of the affected payments (which is really all of them) by examining all stored events rather than the one we just received. No state transition restrictions are enforced.

This brings sandbox into line with GoCardless in terms of its (theoretical) ability to handle out-of-order webhooks.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/webhook/sandbox/resources/WebhookSandboxResource.java
+++ b/src/main/java/uk/gov/pay/directdebit/webhook/sandbox/resources/WebhookSandboxResource.java
@@ -9,12 +9,14 @@ import uk.gov.pay.directdebit.payments.model.Payment;
 import uk.gov.pay.directdebit.payments.model.PaymentState;
 import uk.gov.pay.directdebit.payments.model.SandboxPaymentId;
 import uk.gov.pay.directdebit.payments.services.PaymentService;
+import uk.gov.pay.directdebit.webhook.gocardless.services.WebhookSandboxService;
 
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/pay/directdebit/webhook/sandbox/resources/WebhookSandboxResource.java
```diff
@@ -47,12 +51,13 @@ public Response handleWebhook() {
     }
 
     private void processPayment(Payment payment) {
```
danworth: Wonder whether this is better suited to being in the `WebhookSandboxService` than the resource.

src/test/java/uk/gov/pay/directdebit/events/dao/SandboxEventDaoIT.java
```diff
@@ -59,4 +63,38 @@ public void shouldInsertAnEvent() throws IOException {
         assertThat(foundSandboxEvent.get("event_cause"), is("anEventCause"));
         assertThat((Timestamp) foundSandboxEvent.get("created_at"), isDate(CREATED_AT));
     }
+
+    @Test
+    public void shouldFindLatestApplicableEventForPayment() {
```
danworth: I like the various test fixtures you set up to help assure the logic is right, seems reasonable and complete since we search on `paymentId`, `event_action` and `created_at` 👍

PR #606, opened 2019-07-15
PP-5354 Add DAO for govukpay_event TABLE

stephencdaly:
Add DAO with method for persisting a GovUkPayEvent.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/events/model/GovUkPayEvent.java
+++ b/src/main/java/uk/gov/pay/directdebit/events/model/GovUkPayEvent.java
@@ -1,0 +1,111 @@
+package uk.gov.pay.directdebit.events.model;
+
+import uk.gov.pay.directdebit.mandate.model.Mandate;
+import uk.gov.pay.directdebit.payments.model.Payment;
+
+import java.time.ZonedDateTime;
+import java.util.Objects;
+import java.util.Optional;
+
+public class GovUkPayEvent {
+
+    private final Long id;
+    private final Long mandateId;
+    private final Long paymentId;
+    private final ZonedDateTime eventDate;
+    private final ResourceType resourceType;
+    private final String eventType;
+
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/directdebit/events/model/GovUkPayEvent.java
```diff
@@ -0,0 +1,109 @@
...
+        this.eventDate = eventDate;
+        this.resourceType = resourceType;
+        this.eventType = eventType;
+    }
+
+    public Long getId() {
+        return id;
+    }
+
+    public Long getMandateId() {
```
alexbishop1: Return an `Optional` for this?

src/main/java/uk/gov/pay/directdebit/events/model/GovUkPayEvent.java
```diff
@@ -0,0 +1,109 @@
...
+    private final ResourceType resourceType;
+    private final String eventType;
+
+    public GovUkPayEvent(Long id, Long mandateId, Long paymentId, ZonedDateTime eventDate, ResourceType resourceType, String eventType) {
+        this.id = id;
+        this.mandateId = mandateId;
+        this.paymentId = paymentId;
+        this.eventDate = eventDate;
+        this.resourceType = resourceType;
+        this.eventType = eventType;
```
alexbishop1: Can we use `Objects.requireNonNull(…)` for the fields that should always be populated?

stephencdaly:
We don't have external IDs that I can see for any other event. What would we use it for?

src/main/java/uk/gov/pay/directdebit/events/model/GovUkPayEvent.java
stephencdaly: done

src/main/java/uk/gov/pay/directdebit/events/model/GovUkPayEvent.java
stephencdaly: done

alexbishop1:
You’re right, probably nothing.

PR #608, opened 2019-07-15
PP-5302 Validate external state on mandate search

danworth:
- Remove `EXTERNAL` prefix to `ExternalMandateState` values to aid deserialisation.
- Add `ValidExternalMandateState` annotation with validator.
- Add tests for when `state` parameter is an invalid `ExternalMandateState`.
- Modify `MandateSearchDao` to not add `WHERE clause if the list of internal
mandate states is empty since this results in a binding exception. Add test
to assert this.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/common/exception/validation/ExternalMandateStateValidator.java
+++ b/src/main/java/uk/gov/pay/directdebit/common/exception/validation/ExternalMandateStateValidator.java
@@ -1,0 +1,30 @@
+package uk.gov.pay.directdebit.common.exception.validation;
+
+import uk.gov.pay.directdebit.mandate.api.ExternalMandateState;
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import java.util.Arrays;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class ExternalMandateStateValidator implements ConstraintValidator<ValidExternalMandateState, String> {
+
+    private Set<String> validExternalMandateStateValues;
+
+    @Override
+    public void initialize(ValidExternalMandateState annotation) {
+        validExternalMandateStateValues = Arrays.stream(ExternalMandateState.values())
+                .map(ExternalMandateState::getState)
+                .collect(Collectors.toUnmodifiableSet());
+    }
+
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        if (value == null) {
+            return true;
+        }
+
+        return validExternalMandateStateValues.contains(value);
+    }
+}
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/pay/directdebit/common/exception/validation/ExternalMandateStateValidator.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class ExternalMandateStateValidator implements ConstraintValidator<ValidExternalMandateState, String> {
+
+    private Set<String> validExternalMandateStateValues;
+
+    @Override
+    public void initialize(ValidExternalMandateState annotation) {
+        validExternalMandateStateValues = Arrays.stream(ExternalMandateState.values())
+                .map(ExternalMandateState::getState)
+                .collect(Collectors.toSet());
```
alexbishop1: `Collectors.toUnmodifiableSet()`!

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
```diff
@@ -114,10 +114,10 @@ private SqlStatementAndParameters createSqlQuery(MandateSearchParams params, Str
             sqlParams.put("email", "%" + email + "%");
         });
 
-        params.getExternalMandateState().ifPresent(mandateState -> {
+        if (!params.getInternalStates().isEmpty()) {
```
alexbishop1: I’m not entirely sure about this. It would mean that if someone searches for a non-existent state, they’d get all payments that match their other conditions, rather than nothing, which is what I’d expect.

src/test/java/uk/gov/pay/directdebit/mandate/resources/MandateResourceSearchIT.java
```diff
@@ -112,6 +113,7 @@ public void shouldFindOneMandateUsingAllParams() {
                 .queryParams(params)
                 .get(format("/v1/api/accounts/%s/mandates", gatewayAccountFixture.getExternalId()))
                 .then()
+                .log().body()
```
alexbishop1: Remove this please

danworth: 🤦‍♂

src/main/java/uk/gov/pay/directdebit/common/exception/validation/ExternalMandateStateValidator.java
danworth: Thanks

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateSearchDao.java
danworth: We validate the `MandateSearchParams` created via the end-point so it is not possible for a client to encounter this scenario (searching for an invalid external state). I don't think `MandateSearchDao` is the right place to further validate `MandateSearchParams` so I'm inclined to leave it as-is. From the Dao's perspective it is going to include any internal states into the WHERE clause when they are provided by the `MandateSearchParams`.

alexbishop1: Fair enough. If the API validation should stop it then it’s essentially a bug if it ever happens.

PR #609, opened 2019-07-15
Add columns for granular info to mandates and payments tables

alexbishop1:
Mmm, SQL

### DIFF
--- a/src/main/resources/migrations/00060_alter_tables_mandates_payments_add_columns_state_details_state_details_description.sql
+++ b/src/main/resources/migrations/00060_alter_tables_mandates_payments_add_columns_state_details_state_details_description.sql
@@ -1,0 +1,17 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:alter_table_mandates_state_details
+ALTER TABLE mandates ADD COLUMN state_details VARCHAR(255);
+--rollback ALTER TABLE mandates DROP COLUMN state_details;
+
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/00060_alter_tables_mandates_payments_add_columns_state_details_state_details_description.sql
```diff
@@ -0,0 +1,17 @@
...
+--changeset uk.gov.pay:alter_table_mandates_state_details_description
+ALTER TABLE mandates ADD COLUMN state_details_description VARCHAR(255);
+--rollback ALTER TABLE mandates DROP COLUMN state_details_description;
+
+--changeset uk.gov.pay:alter_table_payments_state_details
+ALTER TABLE payments ADD COLUMN state_details VARCHAR(255);
+--rollback ALTER TABLE payments DROP COLUMN state_details;
+
+--changeset uk.gov.pay:alter_table_payments_state_details_description
+ALTER TABLE payments ADD COLUMN state_details_description VARCHAR(255);
```
danworth: Out of interest why not use type TEXT for these? I assume the 255 is arbitrary?

alexbishop1: Not sure there’s any real reason. `TEXT` is not part of the SQL standard but we have used it in other places.

PR #610, opened 2019-07-15
PP-5442 Update mandate state, cause and description from GC event

alexbishop1:
When processing a GoCardless webhook that changes the state of a mandate, store the cause and description if GoCardless supply them.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/common/model/DirectDebitStateWithDetails.java
+++ b/src/main/java/uk/gov/pay/directdebit/common/model/DirectDebitStateWithDetails.java
@@ -1,0 +1,61 @@
+package uk.gov.pay.directdebit.common.model;
+
+import uk.gov.pay.directdebit.payments.model.DirectDebitState;
+
+import java.util.Objects;
+import java.util.Optional;
+
+public class DirectDebitStateWithDetails<T extends DirectDebitState> {
+
+    private final T state;
+    private final String details;
+    private final String detailsDescription;
+
+    public DirectDebitStateWithDetails(T state, String details, String detailsDescription) {
+        this.state = Objects.requireNonNull(state);
+        this.details = details;
+        this.detailsDescription = detailsDescription;
+    }
+
+    public T getState() {
+        return state;
+    }
+
+    public Optional<String> getDetails() {
+        return Optional.ofNullable(details);
+    }
+
+    public Optional<String> getDetailsDescription() {
+        return Optional.ofNullable(detailsDescription);
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        if (this == other) {
+            return true;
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/pay/directdebit/common/model/DirectDebitStateWithDetails.java
```diff
@@ -0,0 +1,61 @@
+package uk.gov.pay.directdebit.common.model;
+
+import uk.gov.pay.directdebit.payments.model.DirectDebitState;
+
+import java.util.Objects;
+import java.util.Optional;
+
+public class DirectDebitStateWithDetails<T extends DirectDebitState> {
```
danworth: Not sure there is much benefit in having the `DirectDebit` prefix since we're in the DirectDebit project and everything is DirectDebit...

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
```diff
@@ -118,18 +120,24 @@
...
+                                                 @Bind("stateDetailsDescription") String detailsDescription);
 
-    @SqlUpdate("UPDATE mandates m SET state = :state FROM gateway_accounts g WHERE m.payment_provider_id = :providerMandateId " +
+    @SqlUpdate("UPDATE mandates m SET state = :state, state_details = :stateDetails, state_details_description = :stateDetailsDescription " +
+            "FROM gateway_accounts g WHERE m.payment_provider_id = :providerMandateId " +
             "AND g.id = m.gateway_account_id AND g.organisation IS NULL AND g.payment_provider = :provider")
     int updateStateByProviderId(@Bind("provider") PaymentProvider paymentProvider,
                                 @Bind("providerMandateId") PaymentProviderMandateId paymentProviderMandateId,
-                                @Bind("state") MandateState mandateState);
+                                @Bind("state") MandateState mandateState,
```
danworth: Would be nice if its possible to `@BindBean` on the new `DirectDebitStateWithDetails` class to reduce the number of args and avoid deconstructing the contents of `DirectDebitStateWithDetails` prematurely...

src/main/java/uk/gov/pay/directdebit/common/model/DirectDebitStateWithDetails.java
alexbishop1: I was following on from `DirectDebitState` but I take your point

src/main/java/uk/gov/pay/directdebit/mandate/dao/MandateDao.java
alexbishop1: Hmm, didn’t think of that! Might change it later.

PR #611, opened 2019-07-15
PP-5354 Add get latest GovUkPayEvent dao method

stephencdaly:


src/main/java/uk/gov/pay/directdebit/events/dao/mapper/GovUkPayEventMapper.java
```diff
@@ -0,0 +1,33 @@
...
+    @Override
+    public GovUkPayEvent map(ResultSet rs, StatementContext ctx) throws SQLException {
+        var builder = GovUkPayEvent.GovUkPayEventBuilder.aGovUkPayEvent()
+                .withId(rs.getLong("id"))
+                .withResourceType(GovUkPayEvent.ResourceType.valueOf(rs.getString("resource_type")))
+                .withEventDate(ZonedDateTime.ofInstant(
+                        rs.getTimestamp("event_date").toInstant(), ZoneOffset.UTC))
+                .withEventType(GovUkPayEvent.GovUkPayEventType.valueOf(rs.getString("event_type")));
+
+        Optional.ofNullable(rs.getObject("mandate_id"))
```
alexbishop1: Any reason we can’t use `getLong(…)` here?

stephencdaly: Because it returns a primitive long, which can't be nullable

alexbishop1: Fair enough

PR #613, opened 2019-07-15
PP-5442 Update payment state, cause and description from event

alexbishop1:
When processing a webhook that changes the state of a payment, store the cause and description if supplied.


src/main/java/uk/gov/pay/directdebit/payments/dao/PaymentDao.java
```diff
@@ -109,15 +110,21 @@
     Optional<Payment> findPaymentByProviderId(@Bind("provider") PaymentProvider paymentProvider,
                                                                @Bind("providerId") PaymentProviderPaymentId providerId);
 
-    @SqlUpdate("UPDATE payments p SET state = :state FROM mandates m, gateway_accounts g WHERE p.payment_provider_id = :providerPaymentId " +
+    @SqlUpdate("UPDATE payments p SET state = :state, state_details = :stateDetails, state_details_description = :stateDetailsDescription " +
+            "FROM mandates m, gateway_accounts g WHERE p.payment_provider_id = :providerPaymentId " +
             "AND m.id = p.mandate_id AND g.id = m.gateway_account_id AND g.organisation = :goCardlessOrganisationId AND g.payment_provider = :provider")
```
stephencdaly: minor style thing, but do you think it would be nice to add some line breaks between the clauses to make the statement easier to read?

alexbishop1: Yeah, probably. Might have a go at the whole file when I get a chance.

PR #621, opened 2019-07-19
PP-5358 Lookup mandate to calculate state

stephencdaly:
We need to take into account GovUkPayEvents when calculating the state of a mandate. This means we need to look up events by the mandate id rather than the GoCardless business key as a mandate may not have the GoCardless details set.

The main part of this change is to not pass around GoCardlessMandateIdAndOrganisationId and instead look up and pass around the mandate on receipt of a webhook.


src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateCalculator.java
```diff
@@ -35,8 +39,20 @@
         this.goCardlessEventDao = goCardlessEventDao;
     }
 
-    Optional<DirectDebitStateWithDetails<MandateState>> calculate(GoCardlessMandateIdAndOrganisationId goCardlessMandateIdAndOrganisationId) {
-        return goCardlessEventDao.findLatestApplicableEventForMandate(goCardlessMandateIdAndOrganisationId, GOCARDLESS_ACTIONS_THAT_CHANGE_STATE)
+    Optional<DirectDebitStateWithDetails<MandateState>> calculate(Mandate mandate) {
+        Optional<GoCardlessEvent> latestGoCardlessEvent = mandate.getPaymentProviderMandateId()
```
alexbishop1: Maybe `latestApplicableGoCardlessEvent` to make it clear it’s not necessarily the absolute latest?

src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessMandateStateCalculator.java
```diff
@@ -35,8 +39,20 @@
...
+                .map(paymentProviderMandateId -> {
+                    GoCardlessOrganisationId goCardlessOrganisationId = mandate.getGatewayAccount().getOrganisation()
+                            .orElseThrow(() -> new GatewayAccountMissingOrganisationIdException(mandate.getGatewayAccount()));
+
+                    return goCardlessEventDao.findLatestApplicableEventForMandate(
+                            (GoCardlessMandateId) paymentProviderMandateId,
```
alexbishop1: Shame we don’t have a generic type parameter on `Mandate` indicating the type of `PaymentProviderMandateId` so we don’t need casts like this. But that might make code that deals with a mandate without knowing what provider it is more painful. So maybe not.

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
```diff
@@ -63,22 +72,41 @@ public void handleEvents(List<GoCardlessEvent> events) {
...
+                .map(this::toMandateBusinessKey)
                 .flatMap(Optional::stream)
                 .distinct()
+                .map(this::getMandate)
+                .flatMap(Optional::stream)
                 .forEach(goCardlessMandateStateUpdater::updateState);
     }
 
-    private static Optional<GoCardlessMandateIdAndOrganisationId> toGoCardlessMandateIdAndOrganisationId(GoCardlessEvent goCardlessEvent) {
+    private Optional<Pair<GoCardlessMandateId, GoCardlessOrganisationId>> toMandateBusinessKey(GoCardlessEvent goCardlessEvent) {
```
alexbishop1: Really not sure about this method name… reminds me of “Internet is serious business”

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
```diff
@@ -63,22 +72,41 @@ public void handleEvents(List<GoCardlessEvent> events) {
...
         var goCardlessMandateIdAndOrganisationId = goCardlessEvent.getLinksMandate()
-                .map(mandateId -> new GoCardlessMandateIdAndOrganisationId(mandateId, goCardlessEvent.getLinksOrganisation()))
+                .map(mandateId -> Pair.of(mandateId, goCardlessEvent.getLinksOrganisation()))
```
alexbishop1: I think I’d prefer to have a specific type, even if it’s used just once, rather than using a `Pair`

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
```diff
@@ -63,22 +72,41 @@ public void handleEvents(List<GoCardlessEvent> events) {
...
                 .orElseGet(() -> {
                     LOGGER.error("GoCardless event " + goCardlessEvent.getGoCardlessEventId() + " has resource_type mandate but no links.mandate");
                     return null;
                 });
         return Optional.ofNullable(goCardlessMandateIdAndOrganisationId);
     }
 
+    private Optional<Mandate> getMandate(Pair<GoCardlessMandateId, GoCardlessOrganisationId> goCardlessMandateBusinessKey) {
+        Mandate mandate = null;
```
alexbishop1: I think we could avoid this by doing something like:

```java
try {
    Mandate mandate = …;
    return Optional.of(mandate);
} catch (MandateNotFoundException e) {
    LOGGER.error("…");
}

return Optional.empty();
```

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
```diff
@@ -63,22 +72,41 @@ public void handleEvents(List<GoCardlessEvent> events) {
...
+
+        try {
+            mandate = mandateQueryService.findByProviderMandateIdAndOrganisationId(
+                    GOCARDLESS,
+                    goCardlessMandateBusinessKey.getLeft(),
+                    goCardlessMandateBusinessKey.getRight());
+        } catch (MandateNotFoundException e) {
+            LOGGER.error(String.format("Could not update status of GoCardless mandate %s for organisation %s because the mandate was not found",
+                    goCardlessMandateBusinessKey.getLeft(),
+                    goCardlessMandateBusinessKey.getRight()));
```
alexbishop1: If we’re going to use a `Pair`, can we at least extract and name the fields first?

```java
var goCardlessMandateId = goCardlessMandateBusinessKey.getLeft();
var goCardlessOrganisationId = goCardlessMandateBusinessKey.getRight();
try {
    mandate = mandateQueryService.findByProviderMandateIdAndOrganisationId(GOCARDLESS, goCardlessMandateId, goCardlessOrganisationId);
} catch (MandateNotFoundException e) {
    LOGGER.error("…", goCardlessMandateId, goCardlessOrganisationId);
}
```

stephencdaly:
Added requested changes in a new commit

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessServiceTest.java
```diff
@@ -75,17 +82,19 @@ public void shouldStoreEventsWithAnUnhandledResource() {
...
         verify(mockedGoCardlessEventService).storeEvent(goCardlessEvent);
     }
 
     @Test
     public void shouldStoreButNotHandleMandateEventsWithAnUnhandledAction() {
         GoCardlessEvent goCardlessEvent = aGoCardlessEventFixture().withResourceType(MANDATES).withAction("not_handled_again").toEntity();
-
         List<GoCardlessEvent> events = Collections.singletonList(goCardlessEvent);
+
+        when(mockedMandateQueryService.findByGoCardlessMandateIdAndOrganisationId(any(), any())).thenReturn(mock(Mandate.class));
```
alexbishop1: We can do better than `any()` by explicitly creating the `goCardlessEvent` with a specific `GoCardlessMandateId` and a specific `GoCardlessOrganisationId`

stephencdaly: changed

PR #633, opened 2019-07-23
PP-5442 Make GoCardless webhook tests check cause becomes details

alexbishop1:
Update the GoCardless webhook integration tests so they check the `details.cause` of the event is persisted and that it also gets persisted as the `state_details` of the mandate or payment.


alexbishop1:
We do that in the new tests like https://github.com/alphagov/pay-direct-debit-connector/blob/master/src/test/java/uk/gov/pay/directdebit/webhook/gocardless/resources/WebhookGoCardlessResourceOutOfOrderMandateEventsIT.java

PR #634, opened 2019-07-24
PP-5433 Lookup payment to calculate state

stephencdaly:
We need to take into account GovUkPayEvents when calculating the state of a payment. We look up GovUkPayEvents by the payment id rather than the payment provider identifiers as a payment may not have the payment provider details set. To accomplish this we need to have the Payment in the calculator rather than just the payment provider identifiers.

Basically mirrors what was done for mandate event handling.


src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessServiceTest.java
```diff
@@ -215,6 +220,15 @@ public void shouldUpdateStatesForMandatesAndPaymentsAffectedByEvents() {
         when(mockedMandateQueryService.findByGoCardlessMandateIdAndOrganisationId(goCardlessMandateId1, goCardlessOrganisationId2))
                 .thenReturn(mandate3);
 
+        Payment payment1 = mock(Payment.class);
+        Payment payment2 = mock(Payment.class);
```
alexbishop1: These could be declared as fields…

```java
@Mock
private Payment mockedPayment1, mockedPayment2
```

… and then used in tests as needed

PR #636, opened 2019-07-24
PP-5474 Integration tests for GoCardless webhooks

alexbishop1:
Add integration tests for GoCardless webhooks with actions that affect the state of a mandate or payment or cause an email to be sent.

Some of these tests are currently ignored because the appropriate handling behaviour hasn’t been implemented yet.


src/test/java/uk/gov/pay/directdebit/webhook/gocardless/resources/WebhookGoCardlessResourceMandateActionsIT.java
```diff
@@ -0,0 +1,182 @@
...
+        assertThat(mandate.get("state"), is(MandateState.EXPIRED.toString()));
+        assertThat(mandate.get("state_details"), is("mandate_expired"));
+        assertThat(mandate.get("state_details_description"), is("The mandate is being marked as expired, because no payments have been collected " +
+                "against it for the dormancy period of your service user number. If you have access to the mandate reinstation API endpoint, you can use " +
+                "this to attempt to set this mandate up again."));
+    }
+
+    private void postWebhook(String webhookBodyResourceName, String webhookSignature) throws IOException {
+        given().port(testContext.getPort())
+                .body(Resources.toString(Resources.getResource(webhookBodyResourceName), UTF_8))
```
danworth: Was going to suggest you could use Dropwizard's `fixture(webhookBodyResourceName)` (which still uses `UTF_8`) but interestingly that doesn't remove the extra blank line at the end of the fixture files and gives a different signature (perhaps there shouldn't actually be an extra blank line at the end). Seems by default `com.google.common.io.Resources` removes blank lines at the end of a file.

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/resources/WebhookGoCardlessResourceOutOfOrderMandateEventsIT.java
```diff
@@ -119,9 +117,6 @@
     @DropwizardTestContext
     private TestContext testContext;
 
-    @Rule
```
danworth: 👍

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/resources/WebhookGoCardlessResourceOutOfOrderPaymentEventsIT.java
```diff
@@ -82,9 +80,6 @@
     @DropwizardTestContext
     private TestContext testContext;
 
-    @Rule
```
danworth: 👍

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/resources/WebhookGoCardlessResourcePaymentActionsIT.java
```diff
@@ -0,0 +1,219 @@
...
+        mandateFixture.insert(testContext.getJdbi());
+        payerFixture.insert(testContext.getJdbi());
+        paymentFixture.insert(testContext.getJdbi());
+    }
+
+    @Test
+    @Ignore
+    public void customerApprovalDeniedChangesStateToCustomerApprovalDenied() throws IOException {
+        postWebhook("gocardless-webhook-payment-customer-approval-denied.json",
+                "d8040d2583728992037e0c345cc1818c264b192ec62c6a0fd2ac654ac7f4823e");
```
danworth: As per the mandate test I think it would be beneficial to have a helper method generate the signature so we can more easily modify and add new fixtures without having to pre-calculate it and hard-code it (no doubt you could do it in a fashion that its used by both tests).

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/resources/WebhookGoCardlessResourcePaymentActionsIT.java
```diff
@@ -0,0 +1,219 @@
...
+        given().port(testContext.getPort())
+                .body(Resources.toString(Resources.getResource(webhookBodyResourceName), UTF_8))
+                .header("Webhook-Signature", webhookSignature)
+                .accept(APPLICATION_JSON)
+                .post("/v1/webhooks/gocardless")
+                .then()
+                .statusCode(OK.getStatusCode());
+    }
+
+    private void stubEmail(EmailTemplate template) throws JsonProcessingException {
```
danworth: There's some duplication with the Mandate test but since it needs the wiremock rule and its pretty basic I like that it keeps the test easier to read by having it here also.

src/test/resources/gocardless-webhook-mandate-active.json
```diff
@@ -0,0 +1,20 @@
+{
```
danworth: Unless you want me to, I'm not going to review the fixtures in detail as you've no doubt conscientiously created them from the Gocardless docs. However, let me know if you rather I did.

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/resources/WebhookGoCardlessResourceMandateActionsIT.java
```diff
@@ -0,0 +1,182 @@
...
+        gatewayAccountFixture.insert(testContext.getJdbi());
+        mandateFixture.insert(testContext.getJdbi());
+        payerFixture.insert(testContext.getJdbi());
+    }
+
+    @Test
+    @Ignore
+    public void submittedChangesStateToSubmittedToBank() throws IOException {
+        postWebhook("gocardless-webhook-mandate-submitted.json",
+                "d661429adff81d7a7b584eb8f91fece6ee642002a7295db3289eb7fa5da6b750");
```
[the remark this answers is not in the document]
alexbishop1: Yeah, we’re not testing the verification of the signature here (we have `WebhookVerifierTest` for that!).

I did actually start trying to make it so the test calculates the correct signature for the payload but it turned out to be a little more involved than I thought. How about I do that in another PR?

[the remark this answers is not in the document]
danworth: Yeah ok

src/test/resources/gocardless-webhook-mandate-active.json
alexbishop1: These are all real (well, real test) webhooks sent from GoCardless’s sandbox environment, so they should be legit.

PR #640, opened 2019-07-25
PP-5433 Include GovUkPayEvents in Sandbox payment state calculation

stephencdaly:
When calculating the state of a Sandbox payment, find the most recent Sandbox and GovUkPayEvent then use whichever has a later timestamp to calculate the state.

Add ...EventToPaymentStateMapper classes to share the GovUkPayEvent mapping between the GoCardless and Sandbox calculators. Add classes for GoCardless and Sandbox for consistency, as this was done for the mandate state calculation.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/payments/services/gocardless/GoCardlessEventToPaymentStateMapper.java
+++ b/src/main/java/uk/gov/pay/directdebit/payments/services/gocardless/GoCardlessEventToPaymentStateMapper.java
@@ -1,0 +1,28 @@
+package uk.gov.pay.directdebit.payments.services.gocardless;
+
+import uk.gov.pay.directdebit.common.model.DirectDebitStateWithDetails;
+import uk.gov.pay.directdebit.events.model.GoCardlessEvent;
+import uk.gov.pay.directdebit.payments.model.PaymentState;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import static uk.gov.pay.directdebit.payments.model.PaymentState.FAILED;
+import static uk.gov.pay.directdebit.payments.model.PaymentState.PAID_OUT;
+
+public class GoCardlessEventToPaymentStateMapper {
+    private static final Map<String, PaymentState> GOCARDLESS_ACTION_TO_PAYMENT_STATE = Map.of(
+            "failed", FAILED,
+            "paid_out", PAID_OUT
+    );
+
+    static final Set<String> GOCARDLESS_ACTIONS_THAT_CHANGE_PAYMENT_STATE = GOCARDLESS_ACTION_TO_PAYMENT_STATE.keySet();
+
+
+    static Optional<DirectDebitStateWithDetails<PaymentState>> mapGoCardlessEventToPaymentState(GoCardlessEvent goCardlessEvent) {
+        return Optional.ofNullable(GOCARDLESS_ACTION_TO_PAYMENT_STATE.get(goCardlessEvent.getAction()))
+                .map(paymentState -> new DirectDebitStateWithDetails<>(
+                        paymentState, goCardlessEvent.getDetailsCause(), goCardlessEvent.getDetailsDescription()));
+    }
+}
--- a/src/main/java/uk/gov/pay/directdebit/payments/services/sandbox/SandboxEventToPaymentStateMapper.java
+++ b/src/main/java/uk/gov/pay/directdebit/payments/services/sandbox/SandboxEventToPaymentStateMapper.java
@@ -1,0 +1,23 @@
+package uk.gov.pay.directdebit.payments.services.sandbox;
+
+import uk.gov.pay.directdebit.common.model.DirectDebitStateWithDetails;
+import uk.gov.pay.directdebit.events.model.SandboxEvent;
+import uk.gov.pay.directdebit.payments.model.PaymentState;
+import uk.gov.pay.directdebit.webhook.gocardless.services.handlers.GoCardlessPaymentHandler;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+public class SandboxEventToPaymentStateMapper {
+    private static final Map<String, PaymentState> SANDBOX_ACTION_TO_PAYMENT_STATE = Map.of(
+            GoCardlessPaymentHandler.GoCardlessPaymentAction.PAID_OUT.toString(), PaymentState.PAID_OUT
+    );
+
+    static final Set<String> SANDBOX_ACTIONS_THAT_CHANGE_PAYMENT_STATE = SANDBOX_ACTION_TO_PAYMENT_STATE.keySet();
+
+    public static Optional<DirectDebitStateWithDetails<PaymentState>> mapSandboxEventToPaymentState(SandboxEvent sandboxEvent) {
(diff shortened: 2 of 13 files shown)


src/main/java/uk/gov/pay/directdebit/payments/services/gocardless/GoCardlessEventToPaymentStateMapper.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import static uk.gov.pay.directdebit.payments.model.PaymentState.SUCCESS;
+
+public class GoCardlessEventToPaymentStateMapper {
+    private static final Map<String, PaymentState> GOCARDLESS_ACTION_TO_PAYMENT_STATE = Map.of(
+            "failed", PaymentState.FAILED,
```
danworth: Might be nice to have static import for `FAILED` since there's one for `SUCCESS`.

src/main/java/uk/gov/pay/directdebit/payments/services/sandbox/SandboxEventToPaymentStateMapper.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import static uk.gov.pay.directdebit.payments.model.PaymentState.SUCCESS;
+
+public class SandboxEventToPaymentStateMapper {
+    private static final Map<String, PaymentState> SANDBOX_ACTION_TO_PAYMENT_STATE = Map.of(
+            WebhookSandboxResource.SandboxEventAction.PAID_OUT.toString(), SUCCESS
```
danworth: Could have a static import for `PAID_OUT` for brevity.

src/test/java/uk/gov/pay/directdebit/payments/services/sandbox/SandboxPaymentStateCalculatorTest.java
```diff
@@ -16,54 +23,123 @@
...
+import static uk.gov.pay.directdebit.payments.services.sandbox.SandboxEventToPaymentStateMapper.SANDBOX_ACTIONS_THAT_CHANGE_PAYMENT_STATE;
 
-@RunWith(MockitoJUnitRunner.class)
+@RunWith(JUnitParamsRunner.class)
 public class SandboxPaymentStateCalculatorTest {
 
     private static final SandboxPaymentId SANDBOX_PAYMENT_ID = SandboxPaymentId.valueOf("Sandbox payment ID");
 
+    @Rule
+    public MockitoRule mockitoRule = MockitoJUnit.rule();
```
danworth: This doesn't appear to be used anywhere, what is it for?

src/test/java/uk/gov/pay/directdebit/payments/services/sandbox/SandboxPaymentStateCalculatorTest.java
```diff
@@ -16,54 +23,123 @@
...
-        
+
         payment = aPaymentFixture()
                 .withMandateFixture(mandateFixture)
                 .withPaymentProviderId(SANDBOX_PAYMENT_ID)
                 .toEntity();
+    }
+
+    @Test
+    @Parameters({
```
danworth: Wondering if its necessary to use `@Parameters` when there's only one scenario, is this to make it simpler if future states are added?

src/test/java/uk/gov/pay/directdebit/payments/services/sandbox/SandboxPaymentStateCalculatorTest.java
```diff
@@ -16,54 +23,123 @@
...
+            "PAID_OUT, SUCCESS"
+    })
+    public void sandboxEventActionMapsToState(String action, String expectedState) {
+        given(mockSandboxEvent.getEventAction()).willReturn(action);
```
danworth: Out of interest why did you opt to mock the sandbox event rather than create one with the builder. Not saying I think its right or wrong just curious although there is now a mix of creating events and using mocks (see `govUkPayEventTypeMapsToState` which creates a `GovUkPayEvent`.
```
var sandboxEvent = aSandboxEvent()
                .withEventAction(action)
                .build();
```

src/test/java/uk/gov/pay/directdebit/payments/services/sandbox/SandboxPaymentStateCalculatorTest.java
stephencdaly: Yeh I was thinking it will make it a tiny bit easier in the future. I know that we're going to be adding another GovUkPayEvent. We don't have any plans to add another SandboxEvent, so maybe this isn't worthwhile but maybe it is.

src/test/java/uk/gov/pay/directdebit/payments/services/sandbox/SandboxPaymentStateCalculatorTest.java
stephencdaly: Oh yeh I'll do that. I kept the mock behaviour because there was no fixture like there is for other events and wasn't sure it was worthwhile adding one. But as this doesn't need to be inserted into the database I can just use the builder on the SandboxEvent. Good suggestion 👍

src/test/java/uk/gov/pay/directdebit/payments/services/sandbox/SandboxPaymentStateCalculatorTest.java
stephencdaly: It's needed because the class has parameterised tests so uses the parameterised runner rather than the Mockito runner. JUnit5 does this better I think because it gets rid of the idea of runners. We should use JUnit5!

src/main/java/uk/gov/pay/directdebit/payments/services/gocardless/GoCardlessEventToPaymentStateMapper.java
stephencdaly: 👍 done

src/main/java/uk/gov/pay/directdebit/payments/services/sandbox/SandboxEventToPaymentStateMapper.java
stephencdaly: 👍 done

src/test/java/uk/gov/pay/directdebit/payments/services/sandbox/SandboxPaymentStateCalculatorTest.java
stephencdaly: changed

src/test/java/uk/gov/pay/directdebit/payments/services/sandbox/SandboxPaymentStateCalculatorTest.java
danworth: Ah I see, thanks for the explanation

src/test/java/uk/gov/pay/directdebit/payments/services/sandbox/SandboxPaymentStateCalculatorTest.java
danworth: I don't feel strongly, the test is readable as-is. I like the parameterised tests for multiple scenarios as it removes quit a bit of boiler plate.

PR #643, opened 2019-07-25
PP-5478 Unhardcode webhook signatures in webhook integration tests

alexbishop1:
In integration tests that hit the GoCardless webhook endpoint but aren’t trying to test the signature verification code, calculate what the signature should be transparently in the test, rather
than hardcoding the expected value. This makes the tests less brittle because there’s no hardcoded signature to update whenever the test webhook changes.

Separate out the code that calculates a webhook signature from the code that verifies a signature against the expected value in order to facilitate this.


src/main/java/uk/gov/pay/directdebit/webhook/gocardless/support/GoCardlessWebhookVerifier.java
```diff
@@ -0,0 +1,29 @@
...
+import uk.gov.pay.directdebit.webhook.gocardless.exception.InvalidWebhookException;
+
+import static java.lang.String.format;
+
+public class GoCardlessWebhookVerifier {
+    
+    private final GoCardlessWebhookSignatureCalculator goCardlessWebhookSignatureCalculator;
+
+    public GoCardlessWebhookVerifier(String secret) {
+        goCardlessWebhookSignatureCalculator = new GoCardlessWebhookSignatureCalculator(secret);
```
danworth: This doesn't really follow dependency injection with the `new` in the constructor. We can `@Inject` it in and create the binding within `DirectDebitModule` using: 
```
@Provides
    @Singleton
    public GoCardlessWebhookSignatureCalculator provideGoCardlessWebhookSignatureCalculator() {
        if (StringUtils.isBlank(configuration.getGoCardless().getWebhookSecret())) {
            throw new RuntimeException("GoCardless webhook secret is blank");
        }
        return new GoCardlessWebhookSignatureCalculator(configuration.getGoCardless().getWebhookSecret());
    }
```
To check it works before suggesting it I made the changes, if you want to see them I could add a commit or put onto a throw-away branch for your perusal. As part of the change this method can go from `GoCardlessFactory`

```
public GoCardlessWebhookVerifier buildSignatureVerifier() {
        if (StringUtils.isBlank(webhookSecret)) {
            throw new RuntimeException("GoCardless webhook secret is blank");
        }

        return new GoCardlessWebhookVerifier(webhookSecret);
    }
```

alexbishop1: Made this change.

PR #645, opened 2019-07-25
PP-5433 Update payment state after inserting a GovUkPayEvent

stephencdaly:
Recalculate the Payment state and update it if necessary after inserting a GovUkPayEvent. Currently there is only an event for submitted to provider.

### DIFF
--- a/src/test/java/uk/gov/pay/directdebit/payments/services/PaymentServiceTest.java
+++ b/src/test/java/uk/gov/pay/directdebit/payments/services/PaymentServiceTest.java
@@ -4,10 +4,11 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnitRunner;
-import uk.gov.pay.directdebit.app.config.DirectDebitConfig;
 import uk.gov.pay.directdebit.app.config.LinksConfig;
 import uk.gov.pay.directdebit.events.services.DirectDebitEventService;
 import uk.gov.pay.directdebit.events.services.GovUkPayEventService;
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/directdebit/payments/services/PaymentServiceTest.java
```diff
@@ -329,19 +278,20 @@ public void submitPaymentToProvider_shouldSubmitAndRegisterPaymentSubmittedEvent
...
-        assertThat(returnedPayment.getChargeDate(), is(Optional.of(chargeDate)));
-
-        Payment expectedPaymentWithStatePending = fromPayment(returnedPayment).withState(SUBMITTED_TO_PROVIDER).build();
-        verify(mockedDirectDebitEventService).registerPaymentSubmittedToProviderEventFor(expectedPaymentWithStatePending);
+        assertThat(returnedPayment, is(mockPayment));
+
+        Payment submittedPayment = capturedPayment.getValue();
+        assertThat(submittedPayment.getAmount(), is(payment.getAmount()));
+        assertThat(submittedPayment.getMandate(), is(payment.getMandate()));
+        assertThat(submittedPayment.getState(), is(CREATED));
```
alexbishop1: Shouldn’t this be `SUBMITTED_TO_PROVIDER`?

stephencdaly: I'm kind of intercepting the payment before the state is updated to check that the payment service is doing the correct thing before passing it to have the status updated and calculated (which is mocked out).
I think possibly there is some problems with the responsibilities of the class under test that forces me to have to do this, which can hopefully be tidied up as part of the story to remove the DirectDebitEvent stuff.

alexbishop1: Could you name it something other than `submittedPayment`? Since it’s the payment _before_ it’s been submitted.

stephencdaly: it's the payment after it's been submitted but before we've changed the state.

It has the provider id and the charge date.

This can be made better as part of refactoring for the DirectDebitEvent stuff.

alexbishop1: OK, you’ve persuaded me.

PR #666, opened 2019-07-29
PP-5459 Increase minimum amount for payments to 100 pence

danworth:
Gocardless have a minimum of 100 pence for payments. This increases the minimum
amount used for validation and brings it inline with PublicApi and GoCardless.


src/main/java/uk/gov/pay/directdebit/payments/api/CollectPaymentRequestValidator.java
```diff
@@ -13,7 +13,7 @@
     private final static String REFERENCE_KEY = "reference";
     private final static String MANDATE_ID_KEY = "mandate_id";
 
-    private final static int MIN_AMOUNT_IN_PENCE = 1;
+    private final static int MIN_AMOUNT_IN_PENCE = 100;
```
alexbishop1: I suppose really this should be `1_00` for consistency with `MAX_AMOUNT_IN_PENCE`

PR #677, opened 2019-07-30
PP-5451 Call to send email directly in GoCardlessPaymentHandler

stephencdaly:
Remove method in PaymentService to send email for a failed payment as it just directly calls a method in UserNotificationService. Instead use UserNotificationService in GoCardlessPaymentHandler.

Also simplify the handler classes by removing the GoCardlessHandler base class which is now redundant and just implementing the GoCardlessActionHandler interface.


stephencdaly:
Yeh I was thinking it would be nicer to have a switch statement but wasn't sure whether we preferred the existing pattern. Should I go ahead and replace with a switch?

alexbishop1:
I have no real love for the existing pattern, which was very generic and so abstract that it could be hard to follow.

PR #679, opened 2019-07-30
PP-5483 Remove ApacheConnector from our JerseryClient set-up

danworth:
It appears that ApacheConnector does not respect the `http.nonProxyHosts` setting
which is causing problems when DDConnector attempts to call Adminusers (receives
a 403 forbidden from the proxy which it shouldn't be attempting to go via). This
switches our client implementation to the same as PublicApi. There is no setting
in DDConnector config to control whether or not to use SSL (as there is in PublicApi)
and we don't see a need to turn this on or off so it will always set-up SSL.

with @RoryMalcolm


src/main/java/uk/gov/pay/directdebit/notifications/clients/ClientFactory.java
```diff
@@ -1,78 +1,46 @@
...
         this.environment = environment;
         this.conf = conf;
     }
 
-    public Client createWithDropwizardClient(String clientName) {
-        JerseyClientConfiguration clientConfiguration = conf.getClientConfiguration();
-        JerseyClientBuilder defaultClientBuilder = new JerseyClientBuilder(environment)
-                .using(new ApacheConnectorProvider())
-                .using(clientConfiguration)
-                .withProperty(ClientProperties.READ_TIMEOUT, (int) conf.getCustomJerseyClient().getReadTimeout().toMilliseconds())
```
rhowe-gds: We use this to set a 9 second (!) timeout via `config.yaml`

9s seems like an awfully long time, but this will presumably flip us to the dropwizard default of 500ms

RoryMalcolm: We don't have any retry capability atm, in the future we might and can reduce, and while 9 seconds is long, 500ms is probably too short?

rhowe-gds: Looking at `dropwizard-client` it's not clear to me that the read timeout is set by dropwizard at all. It sets 'timeout' on its apache client and calls it the read timeout: https://github.com/dropwizard/dropwizard/blob/master/dropwizard-client/src/main/java/io/dropwizard/client/DropwizardApacheConnector.java#L143

but I don't see an equivalent for the jersey client

rhowe-gds: `HttpClientBuilder` uses the timeout value in two places:

https://github.com/dropwizard/dropwizard/blob/master/dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java#L348
and
https://github.com/dropwizard/dropwizard/blob/master/dropwizard-client/src/main/java/io/dropwizard/client/HttpClientBuilder.java#L355

src/main/java/uk/gov/pay/directdebit/notifications/clients/ClientFactory.java
```diff
@@ -1,78 +1,46 @@
...
+    public Client createWithDropwizardClient() {
+
+        ClientBuilder clientBuilder = ClientBuilder.newBuilder();
 
-        if (System.getProperty(PROXY_HOST_PROPERTY) != null && System.getProperty(PROXY_PORT_PROPERTY) != null) {
-            defaultClientBuilder.withProperty(ClientProperties.PROXY_URI, format("http://%s:%s",
-                    System.getProperty(PROXY_HOST_PROPERTY), System.getProperty(PROXY_PORT_PROPERTY))
-            );
+        try {
+            SSLContext sslContext = SSLContext.getInstance(TLSV1_2);
```
rhowe-gds: I *think* this is the default in Dropwizard: https://github.com/dropwizard/dropwizard/blob/master/dropwizard-client/src/main/java/io/dropwizard/client/ssl/TlsConfiguration.java#L16

danworth: Yeah thanks for pointing that out. I've made some changes to simplify it...

src/main/java/uk/gov/pay/directdebit/notifications/clients/ClientFactory.java
danworth: Now using Dropwizard Jersey Client which can take a `JerseyClientConfiguration` and can specify timeout there...

rhowe-gds: It does, but I'm not convinced it configured this specific timeout. However, the way we configure the timeout currently is explicitly called out in the `dropwizard-client` documentation as unsupported and unlikely to work so I think I'm OK with this change.

rhowe-gds:
`customJerseyClient` in the configuration YAML files can go now as can `CustomJerseyClientConfiguration`

PR #682, opened 2019-07-30
PP-5475 Update mandate external states

stephencdaly:
Update mandate external states to make it more obvious what they mean and the distinction between states that occurred during user setup and after the mandate has been submitted to the payment provider clearer.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/mandate/model/MandateState.java
+++ b/src/main/java/uk/gov/pay/directdebit/mandate/model/MandateState.java
@@ -3,21 +3,22 @@ package uk.gov.pay.directdebit.mandate.model;
 import uk.gov.pay.directdebit.mandate.api.ExternalMandateState;
 import uk.gov.pay.directdebit.payments.model.DirectDebitState;
 
+import static uk.gov.pay.directdebit.mandate.api.ExternalMandateState.ABANDONED;
 import static uk.gov.pay.directdebit.mandate.api.ExternalMandateState.INACTIVE;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/directdebit/mandate/model/MandateState.java
```diff
@@ -3,21 +3,22 @@
...
     CREATED(ExternalMandateState.CREATED),
     AWAITING_DIRECT_DEBIT_DETAILS(STARTED),
-    USER_CANCEL_NOT_ELIGIBLE(ExternalMandateState.CANCELLED),
-    SUBMITTED_TO_PROVIDER(ExternalMandateState.PENDING),
-    SUBMITTED_TO_BANK(ExternalMandateState.PENDING),
+    SUBMITTED_TO_PROVIDER(PENDING),
+    SUBMITTED_TO_BANK(PENDING),
     ACTIVE(ExternalMandateState.ACTIVE),
-    FAILED(INACTIVE),
+    FAILED(ExternalMandateState.FAILED),
```
alexbishop1: Since we’ve going to have to use the long form of the name for this one, we may as well for the others for consistency.

stephencdaly: for payments we've called them EXTERNAL_XXX, is that preferable?

alexbishop1: Yes, let’s do that

PR #683, opened 2019-07-30
PP-5483 Unwrap Optional in Notify Emails

RoryMalcolm:
- Currently emails sent from notify contain optional objects which are
not unwrapped, when Optional.toString() is called on them this then
displays confusing output to the user, eg Optional[YourSunOutputHere],
to fix this the Optional is now fully unwrapped with Optional.get()
beforehand.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
+++ b/src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
@@ -60,7 +60,7 @@ public class UserNotificationService {
             return;
         }
 
-        var personalisation = Map.of(MANDATE_REFERENCE_KEY, mandate.getMandateBankStatementReference().toString(),
+        var personalisation = Map.of(MANDATE_REFERENCE_KEY, mandate.getMandateBankStatementReference().get().toString(),
                 BANK_ACCOUNT_LAST_DIGITS_KEY, mandate.getPayer().orElseThrow(
                         () -> new PayerNotFoundException(mandate.getExternalId())).getAccountNumberLastTwoDigits(),
                 STATEMENT_NAME_KEY, sunName.get().toString(),
@@ -94,7 +94,7 @@ public class UserNotificationService {
 
         var personalisation = Map.of(AMOUNT_KEY, formatToPounds(payment.getAmount()),
                 COLLECTION_DATE_KEY, DATE_TIME_FORMATTER.format(chargeDate),
-                MANDATE_REFERENCE_KEY, mandate.getMandateBankStatementReference().toString(),
+                MANDATE_REFERENCE_KEY, mandate.getMandateBankStatementReference().get().toString(),
                 BANK_ACCOUNT_LAST_DIGITS_KEY, mandate.getPayer().orElseThrow(
                         () -> new PayerNotFoundException(mandate.getExternalId())).getAccountNumberLastTwoDigits(),
                 STATEMENT_NAME_KEY, sunName.get().toString(),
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java
```diff
@@ -60,7 +60,7 @@ public void sendMandateCreatedEmailFor(Mandate mandate) {
             return;
         }
 
-        var personalisation = Map.of(MANDATE_REFERENCE_KEY, mandate.getMandateBankStatementReference().toString(),
+        var personalisation = Map.of(MANDATE_REFERENCE_KEY, mandate.getMandateBankStatementReference().get().toString(),
```
danworth: Perhaps this should handle the case when the optional is empty otherwise `get()` will blow up, whether that be to default it to something or throw an error as for getting the payer next line down. We won't have a bank statement reference if the mandate hasn't been submitted to gocardless as they provide it to us as part of the process. Not sure if we ever send an email prior to that state; if not then maybe ok as-is but it contradicts the idea of optionals to force us to handle the possibility of it being empty and relies on the existing behaviour not changing.

RoryMalcolm: Agree w/ this but there's a big comment further up in the file that talks about why they did it like this elsewhere in the file so I followed convention

danworth: That's fine I understand you're following something already, theres and no need to change it. I don't think the existing is good practice though as that method is relying on some far removed context that it was called within in order to not blow up. Java optionals provide myriad of ways to deal with values being there or not that are succinct and easy to understand. Intelij will also highlight the unconditional `get()` and it might confuse future readers, or those looking at it in Github and not able to see the giant comment a few lines above :)

alexbishop1: I also winced a bit when I saw the `get()`.

I think it’s a bit more subtle than `mandate.getMandateBankStatementReference()` returning an `Optional` that may or may not be present. In this case, we do actually expect it to be present: we should only be sending a mandate setup email if the mandate has been submitted to the provider and therefore has a bank statement reference. If it isn’t, one of our invariants hasn’t held, which a programming issue rather than a data issue.

Long term, I think the answer to this is to encode the state (broad sense of the word) in the type. So we could have two types for a mandate: `UnsubmittedMandate` and `SubmittedMandate`. Only the later would have methods like `getMandateBankStatementReference()`, which would return a non-null `MandateBankStatementReference` rather than an `Optional`. The type system would stop us from calling that getter on an unsubmitted mandate (though we’d probably have a `Mandate` superclass for times where we don’t care precisely which type it is).

When we submit an `UnsubmittedMandate` to the provider, we’d convert the object a `SubmittedMandate`. This would actually be quite easy to do since `Mandate` is already immutable and we create a new object every time it changes.

When we load a mandate from the database, the mapper code could decide whether to instantiate an `UnsubmittedMandate` or a `SubmittedMandate` (or `FailedMandate` or `InactiveMandate`…) by examining the non-null rows returned by the database query. We could even have DAO methods that would query for and return only a `SubmittedMandate`, say, though this could make the error handling a little tricky if it’s hard to distinguish between “there is no mandate with this ID” and “there is a mandate with this ID but it’s not in the state you expect!”.

Anyway, not something we’re going to do immediately but I think it would be worth experimenting with some time.

PR #686, opened 2019-07-30
PP-5455: Refactor handling of GoCardless webhook events

oswaldquek:
Refactoring for the following benefits:
- We can assert that events were handled or not in the WebhookGoCardlessServiceTest
- The code path is much clearer and we can reduce complexity in the dependant
  classes (such as GoCardlessMandateHandler / GoCardlessPaymentHandler) so they
  don't have to deal with figuring out whether the event passed to them is valid
  ior not. They can assume the event is valid
- As a precusor to logging invalid events, we can log them in the
  `handleInvalidEvents` method and not rely on the dependant classes (such as
  GoCardlessMandateHandler / GoCardlessPaymentHandler) to log these. This also
  leads to easier-to-understand unit testing.

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation


src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
```diff
@@ -60,21 +66,48 @@
...
+    private void updateStatesForEvents(List<GoCardlessEvent> events) {
         Map<GoCardlessResourceType, List<GoCardlessEvent>> eventsGroupedByResourceType = events.stream().collect(groupingBy(GoCardlessEvent::getResourceType));
-        updateStateForMandateEvents(eventsGroupedByResourceType.getOrDefault(GoCardlessResourceType.MANDATES, Collections.emptyList()));
-        updateStateForPaymentEvents(eventsGroupedByResourceType.getOrDefault(GoCardlessResourceType.PAYMENTS, Collections.emptyList()));
+        updateStateForMandateEvents(eventsGroupedByResourceType.getOrDefault(MANDATES, Collections.emptyList()));
+        updateStateForPaymentEvents(eventsGroupedByResourceType.getOrDefault(PAYMENTS, Collections.emptyList()));
+    }
+
+    private void handleInvalidEvents(List<GoCardlessEvent> events) {
+//        LOGGER.info("unhandled resource type for event with id {} ", event.getGoCardlessEventId());
```
oswaldquek: proper logging will handled in the next PR

src/main/java/uk/gov/pay/directdebit/events/model/GoCardlessMandateAction.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.pay.directdebit.events.model;
+
+import java.util.Optional;
+
+/**
+ * GoCardless mandate actions
+ *
+ * @see <a href="https://developer.gocardless.com/api-reference/#events-mandate-actions">https://developer.gocardless.com/api-reference/#events-mandate-actions</a>
+ */
+public enum GoCardlessMandateAction {
```
alexbishop1: I’m not really sure this enum needs to exist at all, really. We don’t use it with the state update stuff when handling events; we just use strings.

If this enum does continue to exist, I think it should represent all the possible GoCardless mandate actions, rather than just the ones we want to “handle”. We don’t really have a unified idea of handling an event now that we have the state updates separated from the sending of emails and logging of events.

src/main/java/uk/gov/pay/directdebit/payments/services/sandbox/SandboxEventToPaymentStateMapper.java
```diff
@@ -1,17 +1,17 @@
...
-import uk.gov.pay.directdebit.webhook.gocardless.services.handlers.GoCardlessPaymentHandler;
 
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
 public class SandboxEventToPaymentStateMapper {
     private static final Map<String, PaymentState> SANDBOX_ACTION_TO_PAYMENT_STATE = Map.of(
-            GoCardlessPaymentHandler.GoCardlessPaymentAction.PAID_OUT.toString(), PaymentState.PAID_OUT
+            GoCardlessPaymentAction.PAID_OUT.toString(), PaymentState.PAID_OUT
```
alexbishop1: Something’s gone wrong here… we shouldn’t have GoCardless actions in a mapper used by sandbox!

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
```diff
@@ -60,21 +66,48 @@
 
     public void handleEvents(List<GoCardlessEvent> events) {
         events.forEach(goCardlessService::storeEvent);
+        updateStatesForEvents(events);
+        var mapOfBooleanToListOfEvents = events.stream().collect(partitioningBy(event -> shouldBeHandled(event)));
+        handleValidEvents(mapOfBooleanToListOfEvents.get(true));
+        handleInvalidEvents(mapOfBooleanToListOfEvents.get(false));
```
alexbishop1: Essentially these are events we might send emails for and events we log errors for, right?

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
```diff
@@ -28,10 +29,15 @@
...
 import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.partitioningBy;
+import static uk.gov.pay.directdebit.events.model.GoCardlessResourceType.MANDATES;
+import static uk.gov.pay.directdebit.events.model.GoCardlessResourceType.PAYMENTS;
+import static uk.gov.pay.directdebit.events.model.GoCardlessResourceType.PAYOUTS;
 
 public class WebhookGoCardlessService {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(WebhookGoCardlessService.class);
+    private static final List<GoCardlessResourceType> VALID_RESOURCE_TYPES = List.of(PAYMENTS, MANDATES, PAYOUTS);
```
alexbishop1: This could be a `Set` (or even an `EnumSet`!)

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
```diff
@@ -60,21 +66,48 @@
...
+    private void handleValidEvents(List<GoCardlessEvent> events) {
         events.forEach(event -> {
-            GoCardlessActionHandler handler = getHandlerFor(event.getResourceType());
             LOGGER.info("About to handle event of type: {}, action: {}, resource id: {}",
                     event.getResourceType(),
                     event.getAction(),
                     event.getResourceId());
-            handler.handle(event);
+            
+            if (event.getResourceType() == MANDATES) {
```
alexbishop1: We could do something like the `groupingBy` we do for updating states to avoid having to do this check for every event

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
```diff
@@ -144,20 +177,4 @@ private void updateStateForPaymentEvents(List<GoCardlessEvent> eventsThatAffectP
...
-
-    private GoCardlessActionHandler getHandlerFor(GoCardlessResourceType goCardlessResourceType) {
-        switch (goCardlessResourceType) {
-            case PAYMENTS:
-            case PAYOUTS:
-                return goCardlessPaymentHandler;
-            case MANDATES:
-                return goCardlessMandateHandler;
-            default:
-                return this::logUnknownResourceTypeForEvent;
```
alexbishop1: Does this cause us to lose the logging for resource types we don’t expect to receive (resource types, not actions)?

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
oswaldquek: fixed

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
oswaldquek: This will fixed in the next PR when I implement the "Make receipt of any GoCardless refund event log an error" and "Make receipt of any GoCardless subscription event log an error" acceptance criteria.

src/main/java/uk/gov/pay/directdebit/events/model/GoCardlessMandateAction.java
oswaldquek: | If this enum does continue to exist, I think it should represent all the possible GoCardless mandate actions, rather than just the ones we want to “handle”

Maybe we can rename this to "RelevantGoCardlessMandateActions"?

alexbishop1: I guess what I was getting at is that I don’t think we necessarily need a global list of event actions we care about. Essentially we have:

- events that trigger mandate/payment state updates
- events that send emails
- events that log an error
- maybe other things in the future

… and these can overlap (an event could trigger a state update and send an email). Importantly, there are events that fit in to none of these categories and do nothing (they just get stored).

I’m not even really sure we need `GoCardlessMandateHandler` and friends. We could just do something like this:

```java
events.forEach(goCardlessService::storeEvent);
updateStatesForEvents(events);
sendEmailsForEvents(events);
logErrorsForEvents(events);
```

Each method (which could be in a separate service to keep `WebhookGoCardlessService` light) would do its own thing to decide what events are relevant and what it wants to do with them.

On the specifics of the `GoCardlessMandateAction` enum: if we have it, I think it should represent every possible mandate action, so that we could also use it as a field in `GoCardlessEvent` etc. So I’d be in favour of having a more specific name for what we currently have.

oswaldquek: ok. I've created a GoCardlessEventHandler class. it looks much nicer and I managed to achieve a net deletion of code!

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessService.java
alexbishop1: It probably doesn’t make much difference in this case but generally (particularly with large collections), a `Set` will have faster lookups because it will use hashing (or bit field magic if it’s an `EnumSet`) rather than linearly looking through every element in the list.

oswaldquek: Makes sense. I've deleted this line though, as it's no longer needed anyway!

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/GoCardlessEventHandler.java
```diff
@@ -0,0 +1,89 @@
...
+
+import javax.inject.Inject;
+import java.util.List;
+import java.util.function.Predicate;
+
+import static uk.gov.pay.directdebit.events.model.GoCardlessResourceType.MANDATES;
+import static uk.gov.pay.directdebit.events.model.GoCardlessResourceType.PAYMENTS;
+import static uk.gov.pay.directdebit.events.model.GoCardlessResourceType.PAYOUTS;
+
+public class GoCardlessEventHandler {
```
alexbishop1: Since this class just deals with sending emails triggered by events now, could it have a more specific name?

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/GoCardlessEventHandler.java
```diff
@@ -0,0 +1,90 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(GoCardlessEventHandler.class);
+    
+    private static Predicate<GoCardlessEvent> byValidEventsForMandateResourceTypes =
+            event -> MANDATES.equals(event.getResourceType()) &&
+                    Set.of("created", "submitted", "active", "failed", "cancelled")
```
alexbishop1: Not sure we need to filter by action here: if the methods with the `switch` statements were reordered to do the `switch` before retrieving the mandate/payment, the `switch` would take  care of the filtering for us.

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/GoCardlessEventHandler.java
```diff
@@ -0,0 +1,90 @@
...
+                            .contains(event.getAction().toLowerCase());
```
alexbishop1: Shouldn’t need to do `toLowerCase()`: GoCardless actions are already lower-case.

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/GoCardlessEventHandlerForMandatesTest.java
```diff
@@ -48,43 +56,52 @@
             .withPayerFixture(payerFixture);
 
     private GoCardlessEventFixture goCardlessEventFixture = GoCardlessEventFixture.aGoCardlessEventFixture()
+            .withResourceType(GoCardlessResourceType.MANDATES)
             .withLinksOrganisation(organisationIdentifier);
 
     @Test
     public void handle_onAFailedMandateGoCardlessEvent_shouldRegisterAPayEventAsMandateFailed() {
```
alexbishop1: The name of this test (and others) seem to have drifted apart from what’s actually tested, which is sending emails for certain events.

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/GoCardlessEventHandlerForMandatesTest.java
```diff
@@ -48,43 +56,52 @@
...
         when(mockMandateQueryService.findByGoCardlessMandateIdAndOrganisationId(
                 goCardlessEvent.getLinksMandate().get(), goCardlessEventFixture.getLinksOrganisation()))
                 .thenReturn(mandateFixture.toEntity());
 
-        goCardlessMandateHandler.handle(goCardlessEvent);
+        goCardlessEventHandler.handle(List.of(goCardlessEvent));
 
         verify(userNotificationService).sendMandateFailedEmailFor(mandateFixture.toEntity());
+        verify(mockMandateQueryService).findByGoCardlessMandateIdAndOrganisationId(
+                goCardlessEvent.getLinksMandate().get(), goCardlessEventFixture.getLinksOrganisation());
```
alexbishop1: We shouldn’t need to verify this method call because it’s already stubbed. If it isn’t called, the test should fail anyway.

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/GoCardlessEventHandlerForMandatesTest.java
```diff
@@ -48,43 +56,52 @@
...
+        verify(mockPaymentQueryService, never()).findByGoCardlessPaymentIdAndOrganisationId(any(), any());
```
alexbishop1: I don’t think this is necessary, there’s no real harm if the code does a pointless lookup. Might be worth verifying there’s no more interactions with the `userNotificationService` though.

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessServiceTest.java
```diff
@@ -66,63 +60,24 @@
...
-    }
-
-    @Test
-    public void shouldStoreButNotHandleEventsWithSubscriptionsResource() {
-        GoCardlessEvent goCardlessEvent = aGoCardlessEventFixture().withResourceType(SUBSCRIPTIONS).withAction("created").toEntity();
-
-        List<GoCardlessEvent> events = Collections.singletonList(goCardlessEvent);
-        webhookGoCardlessService.handleEvents(events);
-        verify(mockedGoCardlessEventService).storeEvent(goCardlessEvent);
+        GoCardlessEvent first = aGoCardlessEventFixture().withResourceType(GoCardlessResourceType.REFUNDS).toEntity();
```
alexbishop1: I suppose this test could check events for each of the five resources are stored correctly (maybe using a parameterised test?).

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/GoCardlessEventHandlerTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @InjectMocks
+    private GoCardlessEventHandler goCardlessMandateHandler;
+
+    @Test
+    public void shouldNotHandleEventsWithRefundsResource() {
+        GoCardlessEvent goCardlessEvent = aGoCardlessEventFixture().withResourceType(REFUNDS).withAction("created").toEntity();
+        goCardlessMandateHandler.handle(List.of(goCardlessEvent));
+        verify(mockedMandateQueryService, never()).findByGoCardlessMandateIdAndOrganisationId(any(), any());
+        verify(mockPaymentQueryService, never()).findByGoCardlessPaymentIdAndOrganisationId(any(), any());
```
alexbishop1: I think the important part here is that there’s no interactions with the `UserNotificationService`.

oswaldquek: fixed

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/WebhookGoCardlessServiceTest.java
oswaldquek: fixed

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/GoCardlessEventHandlerForMandatesTest.java
oswaldquek: fixed

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/GoCardlessEventHandlerForMandatesTest.java
oswaldquek: fixed

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/GoCardlessEventHandlerForMandatesTest.java
oswaldquek: fixed

oswaldquek: renamed to SendEmailsForGoCardlessMandateEventsHandlerTest

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/GoCardlessEventHandler.java
oswaldquek: fixed

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/GoCardlessEventHandler.java
oswaldquek: fixed. renamed to SendEmailsForGoCardlessEventsHandler

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/SendEmailsForGoCardlessEventsHandler.java
```diff
@@ -0,0 +1,91 @@
...
+public class SendEmailsForGoCardlessEventsHandler {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SendEmailsForGoCardlessEventsHandler.class);
+    
+    private static Predicate<GoCardlessEvent> byValidEventsForMandateResourceTypes =
+            event -> MANDATES.equals(event.getResourceType());
+    
+    private static Predicate<GoCardlessEvent> byValidEventsForPaymentsAndPayoutsResourceTypes =
+            event -> Set.of(PAYMENTS, PAYOUTS).contains(event.getResourceType()) &&
+                    Set.of("created", "submitted", "confirmed", "failed", "paid_out", "paid")
```
alexbishop1: Much like mandates, I think we don’t need to filter by actions here by putting the `if` statement above the payment lookup in the `handlePayment(…)` method.

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/SendEmailsForGoCardlessEventsHandler.java
```diff
@@ -0,0 +1,91 @@
...
+    @Inject
+    public SendEmailsForGoCardlessEventsHandler(MandateQueryService mandateQueryService,
+                                                UserNotificationService userNotificationService,
+                                                PaymentQueryService paymentQueryService) {
+        this.mandateQueryService = mandateQueryService;
+        this.userNotificationService = userNotificationService;
+        this.paymentQueryService = paymentQueryService;
+    }
+
+    public void handle(List<GoCardlessEvent> events) {
```
alexbishop1: `sendEmails`?

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/SendEmailsForGoCardlessEventsHandler.java
```diff
@@ -0,0 +1,91 @@
...
+        events.stream().filter(byValidEventsForMandateResourceTypes).forEach(e -> handleMandate(e));
+        events.stream().filter(byValidEventsForPaymentsAndPayoutsResourceTypes).forEach(e -> handlePayment(e));
+    }
+
+    private void handlePayment(GoCardlessEvent event) {
```
alexbishop1: `sendEmailForPaymentEvent`?

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/SendEmailsForGoCardlessEventsHandler.java
```diff
@@ -0,0 +1,91 @@
...
+        logEvent(event);
+        GoCardlessPaymentId goCardlessPaymentId = event.getLinksPayment()
+                .orElseThrow(() -> new GoCardlessEventHasNoPaymentIdException(event.getGoCardlessEventId()));
+
+        Payment payment = paymentQueryService.findByGoCardlessPaymentIdAndOrganisationId(goCardlessPaymentId, event.getLinksOrganisation())
+                .orElseThrow(() -> new PaymentNotFoundException(goCardlessPaymentId, event.getLinksOrganisation()));
+
+        if (event.getAction().equalsIgnoreCase("failed")) {
```
alexbishop1: Just `equals` should do it because actions should always be lower-case.

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/SendEmailsForGoCardlessEventsHandler.java
```diff
@@ -0,0 +1,91 @@
...
+            userNotificationService.sendPaymentFailedEmailFor(payment);
+        }
+    }
+
+    private void handleMandate(GoCardlessEvent event) {
+        logEvent(event);
+        switch (event.getAction().toLowerCase()) {
```
alexbishop1: Again, no need to convert to lower-case.

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/SendEmailsForGoCardlessEventsHandler.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+
+    private Mandate getMandate(GoCardlessEvent event) {
+        return mandateQueryService.findByGoCardlessMandateIdAndOrganisationId(
+                    event.getLinksMandate().orElseThrow(() -> new GoCardlessEventHasNoMandateIdException(event.getGoCardlessEventId())),
+                    event.getLinksOrganisation());
+    }
+
+    private static void logEvent(GoCardlessEvent event) {
```
alexbishop1: I think this logging has outlived its usefulness now we don’t handle an event in one go. We should already have logging for persisting events, changing statuses due to events and sending emails because of events.

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/SendEmailsForGoCardlessMandateEventsHandlerTest.java
```diff
@@ -48,43 +53,45 @@
...
 
         when(mockMandateQueryService.findByGoCardlessMandateIdAndOrganisationId(
                 goCardlessEvent.getLinksMandate().get(), goCardlessEventFixture.getLinksOrganisation()))
                 .thenReturn(mandateFixture.toEntity());
 
-        goCardlessMandateHandler.handle(goCardlessEvent);
+        sendEmailsForGoCardlessEventsHandler.handle(List.of(goCardlessEvent));
 
-        verify(userNotificationService).sendMandateFailedEmailFor(mandateFixture.toEntity());
+        verify(userNotificationService, times(1)).sendMandateFailedEmailFor(mandateFixture.toEntity());
```
alexbishop1: Don’t need `times(1)`: it’s the default.

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/SendEmailsForGoCardlessMandateEventsHandlerTest.java
```diff
@@ -48,43 +53,45 @@
...
                 .withLinksOrganisation(GoCardlessOrganisationId.valueOf("does_not_exist"))
                 .withAction("cancelled")
                 .withLinksMandate(null)
-                .toEntity());
+                .toEntity();
 
         thrown.expect(GoCardlessEventHasNoMandateIdException.class);
-        goCardlessMandateHandler.handle(goCardlessEvent);
+        sendEmailsForGoCardlessEventsHandler.handle(List.of(goCardlessEvent));
+        verify(userNotificationService, never()).sendMandateCancelledEmailFor(any());
```
alexbishop1: There’s also `verifyZeroInteractions(userNotificationService);` that could be used here.

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/SendEmailsForGoCardlessPaymentEventsHandlerTest.java
```diff
@@ -17,58 +18,63 @@
...
         GoCardlessEvent goCardlessEvent = goCardlessEventFixture
                 .withLinksOrganisation(GoCardlessOrganisationId.valueOf("does_not_exist"))
                 .withAction("created")
                 .withLinksPayment(null)
                 .toEntity();
 
         thrown.expect(GoCardlessEventHasNoPaymentIdException.class);
-        goCardlessPaymentHandler.handle(goCardlessEvent);
+        sendEmailsForGoCardlessEventsHandler.handle(List.of(goCardlessEvent));
+        verify(mockUserNotificationService, never()).sendPaymentFailedEmailFor(any());
```
alexbishop1: Do we have any tests that check we handle multiple events at once?

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/SendEmailsForGoCardlessEventsHandlerTest.java
```diff
@@ -0,0 +1,58 @@
...
+    public MockitoRule mockitoRule = MockitoJUnit.rule();
+    
+    @Mock
+    private UserNotificationService userNotificationService;
+    
+    @InjectMocks
+    private SendEmailsForGoCardlessEventsHandler goCardlessMandateHandler;
+
+    @Test
+    @Parameters({"REFUNDS", "SUBSCRIPTIONS"})
```
alexbishop1: We will also never send emails for payouts

oswaldquek: fixed

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/SendEmailsForGoCardlessPaymentEventsHandlerTest.java
oswaldquek: i've amended assertEmailIsSentForFailedPaymentGoCardlessEvent to deal with two events instead of one

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/SendEmailsForGoCardlessMandateEventsHandlerTest.java
oswaldquek: nice, `verifyZeroInteractions` is more convenient!

src/test/java/uk/gov/pay/directdebit/webhook/gocardless/services/SendEmailsForGoCardlessMandateEventsHandlerTest.java
oswaldquek: fixed

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/SendEmailsForGoCardlessEventsHandler.java
oswaldquek: ok. deleted.

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/SendEmailsForGoCardlessEventsHandler.java
oswaldquek: personally i don't mind a bit of defensive programming here but i've changed it anyway

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/SendEmailsForGoCardlessEventsHandler.java
```diff
@@ -0,0 +1,77 @@
...
+import static uk.gov.pay.directdebit.events.model.GoCardlessResourceType.PAYMENTS;
+import static uk.gov.pay.directdebit.events.model.GoCardlessResourceType.PAYOUTS;
+
+public class SendEmailsForGoCardlessEventsHandler {
+
+    private static Predicate<GoCardlessEvent> byValidEventsForMandateResourceTypes =
+            event -> MANDATES.equals(event.getResourceType());
+    
+    private static Predicate<GoCardlessEvent> byValidEventsForPaymentsAndPayoutsResourceTypes =
+            event -> Set.of(PAYMENTS, PAYOUTS).contains(event.getResourceType());
```
alexbishop1: We never send emails for payouts, so it doesn’t need to be here.

src/main/java/uk/gov/pay/directdebit/webhook/gocardless/services/handlers/SendEmailsForGoCardlessEventsHandler.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public void sendEmails(List<GoCardlessEvent> events) {
+        events.stream().filter(byValidEventsForMandateResourceTypes).forEach(e -> sendEmailForMandateEvents(e));
+        events.stream().filter(byValidEventsForPaymentsAndPayoutsResourceTypes).forEach(e -> sendEmailForPaymentEvents(e));
+    }
+
+    private void sendEmailForPaymentEvents(GoCardlessEvent event) {
+        GoCardlessPaymentId goCardlessPaymentId = event.getLinksPayment()
+                .orElseThrow(() -> new GoCardlessEventHasNoPaymentIdException(event.getGoCardlessEventId()));
```
alexbishop1: If this was done after we know we have to send an email (similar to mandates with its `getMandate()` method), it would be a bit safer (less chance of an event we don’t care about for email purposes blowing us up) and increase consistency.

PR #688, opened 2019-07-31
PP-5493 Change state to error if problem submitting to provider

alexbishop1:
If there is a problem submitting to a mandate or payment to the provider, record an appropriate event for the mandate or payment and change its state to `SETUP_ERROR` or `PROVIDER_ERROR` as appropriate.


alexbishop1:
Yeah, they’ll 500 (or something), which is no change from the current behaviour. Might want to change that in the future.

Only mandates expire (in the user journey sense) and the effect will be that it ends up in `SETUP_ERROR` or `PROVIDER_ERROR` rather than `USER_SETUP_EXPIRED`, which at leasts provides a little more clarity.

PR #697, opened 2019-08-02
PP-5517 Return payment provider id in search payments response

stephencdaly:
Previously, we were selecting the mandate payment_provider_id in the DAO. Fix the query to instead select the payment payment_provider_id. Add a test for the DAO to check the extracted response is what we expect.


src/test/java/uk/gov/pay/directdebit/payments/dao/PaymentViewDaoIT.java
```diff
@@ -85,7 +88,7 @@ public void shouldReturnOnePaymentViewOnly() {
             MandateFixture mandateFixture = aMandateFixture().withGatewayAccountFixture(gatewayAccountFixture).insert(testContext.getJdbi());
             aPayerFixture()
                     .withMandateId(mandateFixture.getId())
-                    .withName("Joe Bog" + i)
+                    .withName("Joe Bloggs" + i)
```
alexbishop1: 👍

PR #701, opened 2019-08-05
PP-5486: Use the system proxy settings for the Jersey client

rhowe-gds:


src/main/java/uk/gov/pay/directdebit/notifications/clients/ClientFactory.java
```diff
@@ -20,9 +21,9 @@ public ClientFactory(Environment environment, DirectDebitConfig conf) {
     }
 
     public Client createWithDropwizardClient(String name) {
-
         Client client = new JerseyClientBuilder(environment)
                 .using(conf.getClientConfiguration())
+                .using(new SystemDefaultRoutePlanner(null))
```
oswaldquek: why is this needed? connector doesn't seem to use it...

RoryMalcolm: Because connector does not speak to external services and internal services like adminusers, DD connector is in a special situation where its client needs to have both of these functionalities and therefore needs special allowances with regards to nonProxyHosts

PR #703, opened 2019-08-05
PP-5518 Fix searching payments by external state

stephencdaly:
This was broken because PaymentViewSearchParams had its own internal mapping of the states, rather than using a single source of truth.

While fixing this, payment search was re-written to match mandate search code to reduce overhead caused by this being done in different ways.


src/main/java/uk/gov/pay/directdebit/common/model/SearchParams.java
```diff
@@ -1,8 +1,67 @@
...
+
+    public Integer getPage() {
+        return page;
+    }
+
+    public Integer getDisplaySize() {
+        return displaySize;
+    }
+
+    public int getOffset() {
```
danworth: 👍 better place for it on the `SearchParams` object.

src/main/java/uk/gov/pay/directdebit/payments/resources/PaymentSearchResource.java
```diff
@@ -32,27 +35,13 @@ public PaymentSearchResource(PaymentSearchService paymentSearchService) {
...
-            @QueryParam("page") Integer pageNumber,
-            @QueryParam("display_size") Integer displaySize,
-            @QueryParam("from_date") String fromDate,
-            @QueryParam("to_date") String toDate,
-            @QueryParam("reference") String reference,
-            @QueryParam("amount") Long amount,
-            @QueryParam("mandate_id") String mandateId,
-            @QueryParam("state") String state,
+            @PathParam("accountId") GatewayAccount gatewayAccount,
+            @Valid @BeanParam PaymentViewSearchParams paymentViewSearchParams,
```
danworth: Nice alignment with Mandate Search and removal of the manual validation 🌮

src/test/java/uk/gov/pay/directdebit/payments/services/PaymentSearchServiceTest.java
```diff
@@ -189,4 +128,22 @@ public void shouldReturnNoRecords_whenPaginationSetToPage2AndNoDisplaySize() {
         assertThat(paymentViewResponse.getLinksForSearchResult().getNextLink(), is(nullValue()));
         assertThat(paymentViewResponse.getLinksForSearchResult().getSelfLink().getHref().contains("?page=2&display_size=500"), is(true));
     }
+
+
+    @NotNull
+    private List<PaymentResponse> createPaymentResponses(int numberOfPayments) {
+        List<PaymentResponse> paymentViewList = new ArrayList<>();
```
danworth: Another way of doing the same thing using streams, only offering out of interest no need to change.
```
return IntStream.rangeClosed(1, numberOfPayments)
                .mapToObj(i -> aPaymentResponse()
                        .withCreatedDate(createdDate)
                        .withState(new ExternalPaymentStateWithDetails(ExternalPaymentState.EXTERNAL_PENDING, "example_details"))
                        .withReference("Pay reference " + i)
                        .withAmount(1000L + i)
                        .withDescription("This is a description " + i)
                        .withPaymentExternalId(RandomIdGenerator.newId())
                        .build())
                .collect(Collectors.toList());
```

PR #710, opened 2019-08-09
DO NOT MERGE PP-5509 DRAFT change logging to json format

danworth:
Include all existing fields and mdc keys which is the default behaviour.
This is just for draft as proof of concept, it must not be merged until
RE-P have completed the necessary work.


danworth:
No longer needed.

PR #711, opened 2019-08-09
PP-5519 Throw Detailed MandateId Invalid Error

RoryMalcolm:
- We now throw a more detailed error when a mandate Id is invalid as
opposed to the older Downstream System Error.


src/main/java/uk/gov/pay/directdebit/mandate/services/MandateQueryService.java
```diff
@@ -34,7 +35,7 @@ public Mandate findByExternalId(MandateExternalId externalId) {
     public Mandate findByExternalIdAndGatewayAccountExternalId(MandateExternalId mandateExternalId, String gatewayAccountExternalId) {
         return mandateDao
                 .findByExternalIdAndGatewayAccountExternalId(mandateExternalId, gatewayAccountExternalId)
-                .orElseThrow(() -> new MandateNotFoundException(mandateExternalId, gatewayAccountExternalId));
+                .orElseThrow(() -> new MandateIdInvalidException(mandateExternalId, gatewayAccountExternalId));
```
stephencdaly: Is there a good reason to create a new exception type rather than putting the error identifier on the existing exception? See PreconditionFailedException for how we can do this (although I don't think anything extends that exception anymore since we removed one-off direct debits)

PR #718, opened 2019-08-15
PP-5511 Add JSON Formatting To DDConnector Logs

RoryMalcolm:
- Make DD Connector use JSON logging as the format for outputting log
lines.
- Edits the yaml file to facilitate this by adding ```type: logstash-console```


alexbishop1:
Presumably we’re going to wait until infra is ready before merging this?

oswaldquek:
Might be a good idea to JSON log https://github.com/alphagov/pay-direct-debit-connector/blob/master/src/main/java/uk/gov/pay/directdebit/mandate/services/gocardless/GoCardlessService.java#L64-L74 as an example?

PR #731, opened 2019-08-30
Tweak logging for errors validating bank details with GoCardless

alexbishop1:
When there’s a `GoCardlessException` while validating bank account details, don’t log two lines, one with the `GoCardlessException` message immediately, and one without the message when the resultant `InternalServerErrorException`. Instead do all the logging when mapping the `InternalServerErrorException` and enhance this so it always includes the names of and messages from the wrapped exceptions.

We can replace this with something even better when we have structured logging in place.

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/common/exception/InternalServerErrorExceptionMapper.java
+++ b/src/main/java/uk/gov/pay/directdebit/common/exception/InternalServerErrorExceptionMapper.java
@@ -16,7 +16,15 @@ public class InternalServerErrorExceptionMapper implements ExceptionMapper<Inter
 
     @Override
     public Response toResponse(InternalServerErrorException exception) {
-        LOGGER.error(exception.getMessage());
+        StringBuilder sb = new StringBuilder();
+        sb.append("Internal server error");
+        if (exception.getMessage() != null) {
+            sb.append(": ").append(exception.getMessage());
+        }
+        if (exception.getCause() != null) {
+            sb.append(ExceptionSummariser.summarise(exception.getCause()));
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/directdebit/common/exception/InternalServerErrorExceptionMapper.java
```diff
@@ -16,7 +16,15 @@
 
     @Override
     public Response toResponse(InternalServerErrorException exception) {
-        LOGGER.error(exception.getMessage());
+        StringBuilder sb = new StringBuilder();
```
danworth: Would it be simpler to pass the throwable to the logger along with the message we want to print and let it print the stack trace which prints the causes along with the classes and line numbers which might be helpful `LOGGER.error(exception.getMessage(), exception.getCause());` or is that too noisy?

alexbishop1: If we log the stack trace it gets split across multiple lines, which is painful to deal with in Splunk. When we have structured logging, we’ll be able to include stacktraces much more easily and we should revisit this then.

danworth: OK, agree it being split makes it harder but was thinking that avoiding this extra code and having the class/line numbers made it worth it. If you want to go this route whilst we wait then ok 👍

alexbishop1:
Yeah, this is really just a modest improvement until we get structured logging.

PR #765, opened 2019-10-16
PP-5735: Update json logging configuration

oswaldquek:
We can remove the timestamp format as the LogstashConsoleAppenderFactory in
pay-java-commons logs the ISO-8601 formatted timestamp by default.

Example log produced by this configuration:
```
{
  "@timestamp": "2019-10-16T11:07:13.070+01:00",
  "@version": "1",
  "message": "Registering jersey handler with root path prefix: /",
  "logger_name": "io.dropwizard.server.DefaultServerFactory",
  "thread_name": "main",
  "level": "INFO",
  "level_value": 20000,
  "label": "directdebit-connector"
}
```

From local testing I've also verified MDC values are logged and structured
argument values in log lines are logged as a key value pair in the json log.
For example,
    ```
    log.info("This is a log line", StructuredArguments.kv("gateway_account_id","abc"))
    ```
 will produce

```
    {
      "@timestamp": "2019-10-16T11:07:13.070+01:00",
      "@version": "1",
      "message": "This is a log line",
      "gateway_account_id": "abc"
      "logger_name": "io.dropwizard.server.DefaultServerFactory",
      "thread_name": "main",
      "level": "INFO",
      "level_value": 20000,
      "label": "directdebit-connector"
    }
```

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/resources/config/config.yaml
+++ b/src/main/resources/config/config.yaml
@@ -11,11 +11,9 @@ logging:
   appenders:
     - type: logstash-console
       threshold: ALL
-      timeZone: UTC
       target: stdout
-      layout:
-        type: json
-        timestampFormat: "yyyy-MM-dd HH:mm:ss.SSS"
+      customFields:
+        label: "directdebit-connector"
     - type: sentry
       threshold: ERROR
       dsn: ${SENTRY_DSN}
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@
         <hamcrest.version>2.1</hamcrest.version>
         <rest-assured.version>4.1.2</rest-assured.version>
         <mockito.version>3.1.0</mockito.version>
-        <pay-java-commons.version>1.0.20191010084537</pay-java-commons.version>
+        <pay-java-commons.version>1.0.20191016090650</pay-java-commons.version>
         <surefire.version>2.22.2</surefire.version>
     </properties>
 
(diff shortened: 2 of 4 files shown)


src/main/resources/config/config.yaml
```diff
@@ -13,9 +13,10 @@ logging:
       threshold: ALL
       timeZone: UTC
       target: stdout
+      customFields:
+        label: "directdebit-connector"
       layout:
         type: json
```
stephencdaly: I don't think we use the layout at all, the Logstash Encoder logs in JSON by default right?

src/main/resources/config/config.yaml
```diff
@@ -13,9 +13,10 @@ logging:
       threshold: ALL
       timeZone: UTC
```
stephencdaly: do we use this timeZone when creating the logger appender?

oswaldquek: we do but it seems like its UTC by default: https://github.com/dropwizard/dropwizard/blob/master/dropwizard-logging/src/main/java/io/dropwizard/logging/AbstractAppenderFactory.java#L116

src/main/resources/config/config.yaml
oswaldquek: you're right. i've removed the `layout` now.

src/main/resources/config/config.yaml
stephencdaly: But is this only used by the layout builder, which we skip out entirely? We could use this value and set the timeZone on the Logstash encoder, but currently I think we ignore it.

oswaldquek: I don't think it related to the layout builder? Our LogstashConsoleAppenderFactory is a subclass of the AbstractAppenderFactory so the timeZone field is used directly.

stephencdaly: If you dig into the source for `AbstractAppenderFactory`, the timeZone only gets accessed by the `buildLayout` method which will never be called. We only use `AbstractAppenderFactory` for the `wrapAsync` method I think

oswaldquek: ok, i've removed this.

PR #772, opened 2019-12-06
Pp 5893

oswaldquek:
not sure if tokens should be logged yet - awaiting clarification

### DIFF
--- a/src/main/java/uk/gov/pay/directdebit/filters/LoggingMDCRequestFilter.java
+++ b/src/main/java/uk/gov/pay/directdebit/filters/LoggingMDCRequestFilter.java
@@ -1,0 +1,47 @@
+package uk.gov.pay.directdebit.filters;
+
+import org.slf4j.MDC;
+import uk.gov.pay.directdebit.gatewayaccounts.dao.GatewayAccountDao;
+
+import javax.inject.Inject;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/directdebit/filters/LoggingMDCRequestFilter.java
```diff
@@ -0,0 +1,50 @@
...
+                    MDC.put(GATEWAY_ACCOUNT_TYPE, gatewayAccount.getType().name());
+                });
+        
+        getPathParameterFromRequest("mandateExternalId", requestContext)
+                .ifPresent(mandateExternalId -> MDC.put(MANDATE_EXTERNAL_ID, mandateExternalId));
+
+        getPathParameterFromRequest("paymentExternalId", requestContext)
+                .ifPresent(mandateExternalId -> MDC.put(PAYMENT_EXTERNAL_ID, mandateExternalId));
+        
+        getPathParameterFromRequest("token", requestContext)
```
oswaldquek: I decided not to add 'token' as a key in pay-java-commons as it's not common across projects like the charge id or gateway account id is.

stephencdaly: we already have `SECURE_TOKEN` in pay-java-commons which you can use

oswaldquek: oops this comment should have been on line 45. fixed.

PR #852, opened 2020-07-21
PP-6763 Remove Direct Debit smoke tests

narinderchana:
Description:
- Removing Direct Debit smoke tests, so snitches aren't fired for Graphite


narinderchana:
Missed the E2E section in the Jenkins file which are also calling tests.

PR #1008, opened 2021-08-10
PP-8394 fix webhook vulnerability

SandorArpa:
## WHAT YOU DID
- StringUtils.equals() doesn't implement a constant-time algorithm for comparing inputs. As a result, an attacker may be able to run a timing attack that can forge a valid Webhook signature for an arbitrary body.


stephencdaly (approved):
LGTM

